PR #8, opened 2015-07-20
Use rdkafka error string as zend exception message.

plediii:
While working with this library, I found I'm having trouble setting configuration parameters.  When I looked through the code, I thought maybe it would be clearer to simply transmit the rdkafka error message back to the client.


arnaud-lb:
Thanks @plediii

PR #35, opened 2016-03-01
Update README.md

nicolas-t:
contributing


arnaud-lb:
Thank you for making php-rdkafka better!

PR #76, opened 2016-09-21
Windows support for PHP 7

weltling:
In general, just a proper config.w32 and some small fixes was required. The underlaying librkafka hevily relies on C99, thus VC14 only. While it possibly could be ported for VC11 support, it probably doesn't make much sense. Thus not doing 5.6 support.

The snapshots for this PR are available under

http://windows.php.net/downloads/pecl/snaps/rdkafka/2.0.0/

and the dependency libs here

http://windows.php.net/downloads/pecl/deps/

Thanks.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -1,2 +1,2 @@
 rdkafka
-Arnaud Le Blanc [lbarnaud] <arnaud.lb@gmail.com> (lead)
+Arnaud Le Blanc
--- a/config.w32
+++ b/config.w32
@@ -1,13 +1,22 @@
 // $Id$
 // vim:ft=javascript
 
-// If your extension references something external, use ARG_WITH
-// ARG_WITH("rdkafka", "for rdkafka support", "no");
-
-// Otherwise, use ARG_ENABLE
-// ARG_ENABLE("rdkafka", "enable rdkafka support", "no");
+ARG_WITH("rdkafka", "for rdkafka support", "no");
 
 if (PHP_RDKAFKA != "no") {
-	EXTENSION("rdkafka", "rdkafka.c");
+	if (CHECK_LIB("librdkafka.lib", "rdkafka", PHP_RDKAFKA) &&
+		CHECK_HEADER_ADD_INCLUDE("librdkafka/rdkafka.h", "CFLAGS_RDKAFKA")) {
+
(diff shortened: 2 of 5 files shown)


arnaud-lb:
You rock ! :)

How can I test that I don't accidentally break windows builds ? Does travis support windows ?

weltling:
Thanks for merging :)

Travis has no Windows support AFAIK. But a snapshot build can be triggered whenever you ask, it's as easy as feeding the build bot with a tarball. As well, if librdkafka needs to be upgraded, please do not hesitate to ping.

For the releases, you should usually get a message from the build bot. The build logs are being uploaded as well. But also about the compatibility - PHP 7 uses vc14 which is C99 complaint and thus has good compatibility with gcc having C99 by default.

Thanks.

ericnilo:
Hi, can you also build for php 5 (Just hoping)
Thanks!

arnaud-lb:
@weltling, what is the policy for the targeted php versions in the windows builds ? People are trying to use the DLLs with other versions than the 7.1 (c.f. #162), and I don't know what to do :)

nick-zh:
@weltling any update on this? Let me know if we can help with this, thanks in advance!

cmb69:
@arnaud-lb, PHP 7.2+ DLLs are supposed to build. I'll see to it.

cmb69:
@arnaud-lb, I built several versions of librdkafka (0.9.1, 0.11.6, 1.1, 1.2, 1.2.1) on Windows, but all but the first one (which is pretty old) segfault during MSHUTDOWN when `rd_kafka_wait_destroyed()` is called. I have not found out whether that is a build issue, a bug in librdkafka or in this extension, but according to a [comment from the librdkafka developer](https://github.com/edenhill/librdkafka/issues/867#issuecomment-263876506) calling `rd_kafka_wait_destroyed()` is no longer necessary as of 0.9.0, so maybe it's best to only call this function for <= 0.9.0 (or actually, don't call it anymore, since PECL/rdkafka 4 requires 0.11 anyway).

What do you think? Or would you have a suggestion what else to check regarding the segfault?

cc @weltling

nick-zh:
@cmb69 Thanks for the feedback :heart: 
@arnaud-lb I will prepare PRs 3.x and 4.x where we get rid of`rd_kafka_wait_destroyed`
I think it is not worth doing this for anything pre 3.x. Let me know what you think.

PR #80, opened 2016-10-08
Added some missing error constants

ytekeli:


arnaud-lb:
Thanks!

It doesn't seem to build on all librdkafka versions: https://travis-ci.org/arnaud-lb/php-rdkafka/jobs/166088324

Some background on the error constants:

Since 0.9, librdkafka provides a rd_kafka_get_err_descs() function that can be used to get a list of all error constants at runtime. If the function exists, we use it to define the PHP constants at runtime: https://github.com/layersistem/php-rdkafka/blob/4c85dad36b8b27ac69e46a00f69e67ea3a609494/rdkafka.c#L625. If it doesn't, we fallback to defining all constants by hand: https://github.com/layersistem/php-rdkafka/blob/4c85dad36b8b27ac69e46a00f69e67ea3a609494/rdkafka.c#L649.

PR #89, opened 2016-11-16
Correct proto for RdKafka\Conf::set()

RobvH:
The set() call proto had the old syntax, still including passing an instance of conf as the first argument.


arnaud-lb:
Thanks

PR #114, opened 2017-05-12
Added support for the stats callback

nyoung:


arnaud-lb:
Thanks! Sorry for the delay, apparently I forgot about this PR until today.

PR #117, opened 2017-05-24
Adding a temp path in SEARCH_PATH for non-root compilation

sandvige:
Since it's not possible to use sudo command on container based image on Travis CI, we're installing librdkafka in /tmp. It's also not possible to tell pecl to use an alternative source of include and library path, this is why we need to have a tmp path in the SEARCH_PATH

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -6,7 +6,7 @@ PHP_ARG_WITH(rdkafka, for rdkafka support,
 
 if test "$PHP_RDKAFKA" != "no"; then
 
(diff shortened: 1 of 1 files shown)


arnaud-lb:
Hi @sandvige 

I believe that `./configure --with-rdkafka=/tmp/librdkafka-staging/usr/local` may work without this change :) If it doesn't, we should fix this.

nick-zh:
@sandvige can we close this?

PR #132, opened 2017-09-26
message : expose len of rdkafka-message

ItzaPhenix:
Hello,

We currently use your API with some php5 worker to consume very high throughput of Kafka message.
We need to get the length payload for each message but we encounter some perf issues to apply strlen on each string payload (and we cannot pass to php7 unfortunatly..).

php-rdkafka doesn't expose len value of the rdkafka-message and force to
apply strlen on the generate string into the php code to retrieve it.
Save this value into the memory will optimize the php client.

I'm not very familiar with Zend api but i think it can fix it.

Regards,


arnaud-lb:
Thanks!

Issue #75, opened 2016-09-20
How to close a consumer after consuming


t3chboy:
How do i close the consumer after subscribing it.  
Following is how i am subscribing it.
Also consumerstop is function of this class RdKafka\ConsumerTopic  so can i use it in below code,because it need partition id as parameter for consumerstop() and while consuming i am not knwing from which partition i m consuming.

```
$consumer = new \RdKafka\KafkaConsumer($conf);
// Subscribe to topic
$consumer->subscribe(['register1']);
$message = $consumer->consume(100);
```

arnaud-lb:
Hi @t3chboy 

Using the unsubscribe function should do what you want: https://arnaud-lb.github.io/php-rdkafka/phpdoc/rdkafka-kafkaconsumer.unsubscribe.html

If you just want to unsubscribe from one topic, call subscribe() with all the topics you want to consume, but the one you don't want to consume anymore.

TiMESPLiNTER:
@arnaud-lb `unsubscribe` method never calls the function [`rd_kafka_consumer_close`](https://github.com/edenhill/librdkafka/blob/52c2005eb323b5876646ac0136caa12e225fdb07/src/rdkafka.c#L2109).

So I have the feeling that a `RdKafkaConsumer` never closes connection properly by calling `unsubscribe`. The only position I've found a call to this method is in the `STORE_OBJECT` call: https://github.com/arnaud-lb/php-rdkafka/blob/7f8dc9a7da0cc770902c157cb374efbacc067446/kafka_consumer.c#L55

Can you explain the reason for this. I'm not that experienced with PHP extension code so I can't really figure out when this function gets called but what I've read so far only a call to this method will close the connection from the consumer to the broker.

My consumers are idling pretty long on the broker from what I can see in the logs. Even after I `unset()` the consumer.

arnaud-lb:
Hi @TiMESPLiNTER 

rd_kafka_consumer_close is called when the RdKafka\KafkaConsumer() instance is destroyed. The instance is destroyed when it's not referenced anymore, either by user code or by other php-rdkafka object instance like a RdKafka\Topic.

TiMESPLiNTER:
@arnaud-lb thanks for your response. Is there a specific reason why there is no `close` method available for the `RdKafka\KafkaConsumer`? I guess my consumers never disconnect because somewhere there's still a reference pointing to it. So it would be great to be able to "force" close it with a method call.

PR #144, opened 2018-01-19
Add missing close method

TiMESPLiNTER:
Adds the possibility to explicitly close the connection between the consumer and the broker.

So far it only happens on complete dereference of consumer object which is not always possible.

See #75

### DIFF
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -530,7 +530,7 @@ PHP_METHOD(RdKafka__KafkaConsumer, close)
     }
 
     rd_kafka_consumer_close(intern->rk);
-    intern->rk = null;
+    intern->rk = NULL;
 }
 /* }}} */
 
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -522,6 +522,8 @@ ZEND_END_ARG_INFO()
 
 PHP_METHOD(RdKafka__KafkaConsumer, close)
 {
+    object_intern *intern;
+
     intern = get_object(getThis() TSRMLS_CC);
     if (!intern) {
         return;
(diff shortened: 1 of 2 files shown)


kravchenko-a:
@arnaud-lb When you merge this feature?

nick-zh:
@arnaud-lb any update on this?

arnaud-lb:
Hi @TiMESPLiNTER 

Thank you for this contribution. What would be the use-case of calling `close()` ? What would happen to inflight / unsent messages ?

I think that we should change `close()` such that it waits for all messages to be sent, before returning. Users could call `purge()` (#211) before `close()`, as needed.

kafka_consumer.c
```diff
@@ -514,6 +514,25 @@ PHP_METHOD(RdKafka__KafkaConsumer, commitAsync)
...
+PHP_METHOD(RdKafka__KafkaConsumer, close)
+{
+    object_intern *intern;
+
+    intern = get_object(getThis() TSRMLS_CC);
+    if (!intern) {
+        return;
+    }
+
+    rd_kafka_consumer_close(intern->rk);
```
arnaud-lb, line 532: We should also set `intern->rk` to null, else it might sill be used by other methods, or by `rd_kafka_free()`.

nick-zh:
@arnaud-lb i think this should be ok, right? Since we have purge and there's an open PR for flush as well

arnaud-lb:
Thanks !

kravchenko-a:
@TiMESPLiNTER you could add usage examples of new method to the documentation.

nick-zh:
@modder2 i added a pr for this

PR #156, opened 2018-03-11
Drop unmaintained PHP versions from travis build and add 7.2

tPl0ch:
PHP versions `5.4` and `5.5` have officially reached EOL, we can safely drop them from the travis build.

Additionally we should start building against `7.2`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
 sudo: true
 language: php
 php:
-    - 5.4
-    - 5.5
     - 5.6
     - 7.0
     - 7.1
+    - 7.2
 env:
     - LIBRDKAFKA_VERSION=0.8.6
     - LIBRDKAFKA_VERSION=0.8


arnaud-lb:
Hi @tPl0ch 

These versions don't seem to incur a noticeable maintenance cost yet. Shouldn't we keep them a bit longer ?

tPl0ch:
@arnaud-lb While you are right from a local maintenance perspective, we should also look at it from an ecosystem perspective. We are giving wrong incentives to people using outdated PHP versions, and as library maintainers we have a great lever to nudge people towards upgrading to maintained versions.

This is also a risk indicator for clients when planning for integration.

simPod:
Also, 7.3 can be added now

simPod:
7.0 EOLed, 5.6 in 12 days

arnaud-lb:
Thanks

PR #193, opened 2018-11-12
Added ports to brokers in usage example.

dbakiu:
The example did not work without the ports. After adding the default broker ports, it worked.


arnaud-lb (approved):

arnaud-lb:
Thanks

Issue #197, opened 2018-12-28
Headers message support


padlavoine:
Hi, 

Is header support in a message provided ?
https://github.com/edenhill/librdkafka/blob/v0.11.5/src/rdkafka.h#L962

Thank you.

* PHP version: >7.1
 * librdkafka version 0.11.5:
 * php-rdkfaka version: 3.5

nick-zh:
It should be now with the newest release of 3.1.0 which was released 5 days ago
I think this can be closed now @padlavoine @arnaud-lb ?

PR #206, opened 2019-04-10
Message headers support

martynaszaliaduonis:
### Message headers support

**Requirements:**
- [librdkafka](https://github.com/edenhill/librdkafka) >= 0.11.4
- Kafka >= 0.11.0.0

**Usage:**
_Produce:_ you need to use `producev` function instead of `produce`.

> producev(int $partition, int $msgflags[, string $payload, string $key, array $headers])

_Consume:_ just use `$message->headers` to retrieve all message headers.

**Structure:**
`producev` accepts only associative array (or null):

> ['key1' => 'value1', 'key2' => 'value2']

`$message->headers` returns associative array or null if no headers.

Requested in #197

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -63,6 +63,12 @@ if test "$PHP_RDKAFKA" != "no"; then
     AC_MSG_WARN([no rd_kafka_message_timestamp, timestamp support will not be available])
   ])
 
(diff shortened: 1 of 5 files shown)


arnaud-lb:
Thanks!

arnaud-lb:
Thanks!

PR #207, opened 2019-04-16
Parse headers only if message don't have error

dariuskasiulevicius:
Some time this method is called with empty message and if we will try parse headers we will get segmentation fault


arnaud-lb:
Thanks!

PR #216, opened 2019-04-29
remove deprecated configs from readme

nick-zh:
I took the liberty to remove deprecated config vars from the doc, to discourage their usage.
If you prefer to still have them, you can of course decline my PR :smile:


nick-zh:
@arnaud-lb thoughts?

arnaud-lb:
Thanks !

Issue #78, opened 2016-09-29
Find last stored commited offset.


t3chboy:
Hi,

```
        $conf->set('group.id','commonqueue2');
        $conf->set('offset.store.method', 'broker');

        $rk = new \RdKafka\Consumer($conf);
        $rk->addBrokers($kafka_servers);

        $topicConf = new \RdKafka\TopicConf();

        $topicConf->set('auto.commit.interval.ms',1000);
        $topicConf->set('offset.store.method', 'broker');
        $topicConf->set('auto.offset.reset', 'smallest');

        $topic = $rk->newTopic("registration17", $topicConf);

        $topic->consumeStart(0,RD_KAFKA_OFFSET_STORED); 

        while (true) {

            $message = $topic->consume(0,1000);

            //store offset to broker
            $topic->offsetStore($message->partition,$message->offset);

            $message_offset = $message->offset;
            echo $message_data = $message->payload;

        }//end of while loop
```

In the above program how can i find the last stored offset before executing following code
`$topic->consumeStart(0,RD_KAFKA_OFFSET_STORED);    
`

```
PHP version: PHP 5.5.9-1ubuntu4.19
librdkafka version: 0.9.1
php-rdkfaka version: 0.9.1-php5
```

t3chboy:
Hi @arnaud-lb  any updates for above issue,i want to implement this but got stuck due to above issue.
Kindly help.

t3chboy:
Hey @arnaud-lb  is there some method in here like position() method in java to get current  offset position.

arnaud-lb:
Hi @t3chboy 

librdkafka has a `rd_kafka_committed()` function, but it's not implemented in php-rdkafka yet.

I think it could be added as `RdKafka\Consumer::committed([]RdKafka\TopicPartition $partitions, $timeout_ms)`

t3chboy:
@arnaud-lb  but how to use it, if i am using it as following, i get error

```
$rk = new \RdKafka\Consumer($conf);
$rd = new \RdKafka\TopicPartition("registration22", 0);
$a = $rk->committed($rd, 100);
print_r($a);
```

or

```
$rk = new \RdKafka\Consumer($conf);
$rd = new \RdKafka\TopicPartition("registration22", 0);
$a = $rk->committed(0 100);
print_r($a);
```

Error : Segmentation fault (core dumped);

remizyaka:
Hey @tPl0ch 
[...]

PR #219, opened 2019-05-08
Expose rd_kafka_query_watermark_offsets

gytislakavicius:
I don't really like returning low and high offsets by reference, but I wanted the extension to maintain the same behaviour as librdkafka.


arnaud-lb:
Thanks!

nick-zh:
https://github.com/arnaud-lb/php-rdkafka/issues/78

PR #222, opened 2019-05-09
Fix KafkaTopic::producev causing segfault on librdkafka 1.0.0

### LINKED ISSUE #221: Segmentation fault using `producev`
* PHP version: 7.1.29
 * librdkafka version: master and v1.0.0
 * php-rdkfaka version: master and 3.1.0

Doing a basic usage, if I do
```php
$topic->produce(RD_KAFKA_PARTITION_UA, 0, 'some payload');
```

It works ok, but if instead I try to use with headers:
```php
$topic->producev(RD_KAFKA_PARTITION_UA, 0, "some payload", null, ['key' => 'value']);
```

I get a `segmentation fault`.

I read about compiling librdkafka with the `--disable-ssl --disable-sasl` options, but also I get a segfault.

Am I doing something wrong?

Steveb-p:
Fixes #221

### DIFF
--- a/topic.c
+++ b/topic.c
@@ -423,6 +423,7 @@ PHP_METHOD(RdKafka__ProducerTopic, producev)
     arglen_t key_len = 0;
     rd_kafka_resp_err_t err;
     kafka_topic_object *intern;
+    kafka_object *kafka_intern;
     HashTable *headersParam = NULL;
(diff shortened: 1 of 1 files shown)


nick-zh:
@arnaud-lb

arnaud-lb:
Thanks!

PR #223, opened 2019-05-10
Run integration tests in CI

### LINKED ISSUE #220: Compatibility with 1.0.0
* PHP version: 7.3
 * librdkafka version: 1.0.0
 * php-rdkafka version: current master (`4f97323bee66519956635c636a9d77a2564839b6`)

I've run tests to prepare myself to give a try to add topic API to phprdkafka and I've noticed that integration tests fail on librdkafka 1.0.0, while they work just fine against 0.11.6. This affects integration tests only, hence probably why they went undetected in regard to CI, since travis seems to run non-integration tests only.

Here is output of my 1.0.0 run:
```shell
[steveb@y700 php-rdkafka]$ make test

Build complete.
Don't forget to run 'make test'.


=====================================================================
PHP         : /usr/bin/php 
PHP_SAPI    : cli
PHP_VERSION : 7.3.4
ZEND_VERSION: 3.3.4
PHP_OS      : Linux - Linux y700 4.19.36-1-MANJARO #1 SMP PREEMPT Sat Apr 20 22:05:46 UTC 2019 x86_64
INI actual  : /home/steveb/CLionProjects/php-rdkafka/tmp-php.ini
More .INIs  :   
CWD         : /home/steveb/CLionProjects/php-rdkafka
Extra dirs  : 
VALGRIND    : Not used
=====================================================================
TIME START 2019-05-09 01:12:56
=====================================================================
FAIL Allow null payload [tests/allow_null_payload.phpt] 
PASS Allow null payload [tests/allow_null_payload_and_key.phpt] 
FAIL Bug 115 [tests/bug115.phpt] 
PASS Bug 74 [tests/bug74.phpt] 
PASS Bug 88 [tests/bug88.phpt] 
PASS RdKafka\Conf [tests/bugConfSetArgument.phpt] 
PASS RdKafka\Conf [tests/conf.phpt] 
PASS RdKafka\Conf [tests/conf_callbacks.phpt] 
FAIL RdKafka\Conf [tests/conf_callbacks_integration.phpt] 
PASS RdKafka\Conf::setDefaultTopicConf() [tests/conf_setDefaultTopicConf.phpt] 
PASS constants [tests/constants.phpt] 
FAIL Message headers [tests/message_headers.phpt] 
FAIL Produce, consume [tests/produce_consume.phpt] 
FAIL Produce, consume queue [tests/produce_consume_queue.phpt] 
PASS rd_kafka_get_err_descs() [tests/rd_kafka_get_err_descs.phpt] 
PASS test0 [tests/test0.phpt] 
PASS RdKafka\TopicConf [tests/topic_conf.phpt] 
PASS RdKafka\TopicPartition [tests/topic_partition.phpt] 
=====================================================================
TIME END 2019-05-09 01:17:06

=====================================================================
TEST RESULT SUMMARY
---------------------------------------------------------------------
Exts skipped    :    0
Exts tested     :   29
---------------------------------------------------------------------

Number of tests :   18                18
Tests skipped   :    0 (  0.0%) --------
Tests warned    :    0 (  0.0%) (  0.0%)
Tests failed    :    6 ( 33.3%) ( 33.3%)
Expected fail   :    0 (  0.0%) (  0.0%)
Tests passed    :   12 ( 66.7%) ( 66.7%)
---------------------------------------------------------------------
Time taken      :  250 seconds
=====================================================================

=====================================================================
FAILED TEST SUMMARY
---------------------------------------------------------------------
Allow null payload [tests/allow_null_payload.phpt]
Bug 115 [tests/bug115.phpt]
RdKafka\Conf [tests/conf_callbacks_integration.phpt]
Message headers [tests/message_headers.phpt]
Produce, consume [tests/produce_consume.phpt]
Produce, consume queue [tests/produce_consume_queue.phpt]
=====================================================================

```
I've double checked that tests run correct librdkafka version (output via `phpinfo()` in `.phpt` file:
```
rdkafka

rdkafka support => enabled
version => 3.1.0-dev
build date => May  9 2019 03:12:16
librdkafka version (runtime) => 1.0.0
librdkafka version (build) => 1.0.0.0

```

When looking into test output, I have for example in `bug115.out`:
```
** ERROR: process timed out **
```
I can see in my Kafka instance log that some communication occurs and it roughly resembles what I've seen for 0.11.6 librdkafka.

For reference that my Kafka is configured properly here's my results for `0.11.6`:
```shell
[steveb@y700 php-rdkafka]$ make test

Build complete.
Don't forget to run 'make test'.


=====================================================================
PHP         : /usr/bin/php 
PHP_SAPI    : cli
PHP_VERSION : 7.3.4
ZEND_VERSION: 3.3.4
PHP_OS      : Linux - Linux y700 4.19.36-1-MANJARO #1 SMP PREEMPT Sat Apr 20 22:05:46 UTC 2019 x86_64
INI actual  : /home/steveb/CLionProjects/php-rdkafka/tmp-php.ini
More .INIs  :   
CWD         : /home/steveb/CLionProjects/php-rdkafka
Extra dirs  : 
VALGRIND    : Not used
=====================================================================
TIME START 2019-05-09 01:30:43
=====================================================================
PASS Allow null payload [tests/allow_null_payload.phpt] 
PASS Allow null payload [tests/allow_null_payload_and_key.phpt] 
PASS Bug 115 [tests/bug115.phpt] 
PASS Bug 74 [tests/bug74.phpt] 
PASS Bug 88 [tests/bug88.phpt] 
PASS RdKafka\Conf [tests/bugConfSetArgument.phpt] 
PASS RdKafka\Conf [tests/conf.phpt] 
PASS RdKafka\Conf [tests/conf_callbacks.phpt] 
PASS RdKafka\Conf [tests/conf_callbacks_integration.phpt] 
PASS RdKafka\Conf::setDefaultTopicConf() [tests/conf_setDefaultTopicConf.phpt] 
PASS constants [tests/constants.phpt] 
PASS Message headers [tests/message_headers.phpt] 
PASS Produce, consume [tests/produce_consume.phpt] 
PASS Produce, consume queue [tests/produce_consume_queue.phpt] 
PASS rd_kafka_get_err_descs() [tests/rd_kafka_get_err_descs.phpt] 
PASS test0 [tests/test0.phpt] 
PASS RdKafka\TopicConf [tests/topic_conf.phpt] 
PASS RdKafka\TopicPartition [tests/topic_partition.phpt] 
=====================================================================
TIME END 2019-05-09 01:31:02

=====================================================================
TEST RESULT SUMMARY
---------------------------------------------------------------------
Exts skipped    :    0
Exts tested     :   29
---------------------------------------------------------------------

Number of tests :   18                18
Tests skipped   :    0 (  0.0%) --------
Tests warned    :    0 (  0.0%) (  0.0%)
Tests failed    :    0 (  0.0%) (  0.0%)
Expected fail   :    0 (  0.0%) (  0.0%)
Tests passed    :   18 (100.0%) (100.0%)
---------------------------------------------------------------------
Time taken      :   19 seconds
=====================================================================
```
And exerpt from `phpinfo()`:
```
rdkafka

rdkafka support => enabled
version => 3.1.0-dev
build date => May  9 2019 03:29:44
librdkafka version (runtime) => 0.11.6
librdkafka version (build) => 0.11.6.0
```

Steveb-p:
Will resolve #220 eventually.

Contains resolution for segfault issue as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,27 @@ php:
     - 7.1
     - 7.2
     - 7.3
-    - 7.4snapshot
+    - 7.4
     - nightly
 matrix:
     allow_failures:
         - php: nightly
 env:
-    - LIBRDKAFKA_VERSION=0.11.x
-    - LIBRDKAFKA_VERSION=v1.0.1
-    - LIBRDKAFKA_VERSION=v1.1.0
-    - LIBRDKAFKA_VERSION=master
-script: ./.travis/build.sh
+    global:
+        - TEST_KAFKA_BROKERS=localhost:9092
+        - TEST_KAFKA_BROKER_VERSION=2.3
+        - LIBRDKAFKA_REPOSITORY_URL=https://github.com/edenhill/librdkafka.git
+        - LIBRDKAFKA_VERSION=v1.2.2
+    matrix:
+        - LIBRDKAFKA_VERSION=0.11.x
+        - LIBRDKAFKA_VERSION=v1.0.1
+        - LIBRDKAFKA_VERSION=v1.1.0
+        - LIBRDKAFKA_VERSION=v1.2.2
+        - LIBRDKAFKA_VERSION=master
+before_script: 
+    - ./.travis/start-kafka.sh
+script: 
+    - ./.travis/build.sh
+    ### Issues with environmental variables causes this copy to be required
+    - cp tests/test_env.php.sample tests/test_env.php
+    - ./.travis/test.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,4 +10,13 @@ Pull requests should be made against the master branch, which supports both PHP 
 
 ## Testing
 
-Tests are in phpt format in the tests directory. They can be run by executing `make test`.
+Tests are in phpt file format in the tests directory.
+
+### Using your own machine for building and testing. 
+
+Tests can be run by following compilation and installation procedure 
+and executing `make test`.
+
+To run integration tests, make sure you have Kafka instance running.
+Then, rename `test_env.php.sample` to `test_env.php` and adjust it
+with values proper for your kafka instance.
--- a/tests/allow_null_payload_and_key.phpt
+++ b/tests/allow_null_payload_and_key.phpt
@@ -2,16 +2,15 @@
 Allow null payload
 --SKIPIF--
 <?php
-file_exists(__DIR__."/test_env.php") || die("skip");
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $topicName = sprintf('test_rdkafka_%s', uniqid());
 
 $producer = new RdKafka\Producer();
-$producer->addBrokers(TEST_KAFKA_BROKERS);
+$producer->addBrokers(getenv('TEST_KAFKA_BROKERS'));
 $topic = $producer->newTopic($topicName);
 
 $topic->produce(0, 0);
@@ -21,7 +20,7 @@ while ($producer->getOutQLen() > 0) {
 }
 
 $consumer = new RdKafka\Consumer();
-$consumer->addBrokers(TEST_KAFKA_BROKERS);
+$consumer->addBrokers(getenv('TEST_KAFKA_BROKERS'));
 
 $topic = $consumer->newTopic($topicName);
 $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
@@ -31,11 +30,11 @@ while (true) {
     if ($message === null) {
         continue;
     }
-    switch ($message->err) {
-        case RD_KAFKA_RESP_ERR_NO_ERROR:
-            var_dump($message->payload);
-            var_dump($message->key);
-            break 2;
+
+    if (RD_KAFKA_RESP_ERR_NO_ERROR === $message->err) {
+        var_dump($message->payload);
+        var_dump($message->key);
+        break;
     }
 }
 
--- a/tests/bug115.phpt
+++ b/tests/bug115.phpt
@@ -2,17 +2,16 @@
 Bug 115
 --SKIPIF--
 <?php
-file_exists(__DIR__."/test_env.php") || die("skip");
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $delivered = 0;
 
 $conf = new RdKafka\Conf();
-if (RD_KAFKA_VERSION >= 0x090000) {
-    $conf->set('broker.version.fallback', TEST_KAFKA_BROKER_VERSION);
+if (RD_KAFKA_VERSION >= 0x090000 && false !== getenv('TEST_KAFKA_BROKER_VERSION')) {
+    $conf->set('broker.version.fallback', getenv('TEST_KAFKA_BROKER_VERSION'));
 }
 $conf->setErrorCb(function ($producer, $err, $errstr) {
     printf("%s: %s\n", rd_kafka_err2str($err), $errstr);
@@ -28,18 +27,18 @@ $conf->setDrMsgCb(function ($producer, $msg) use (&$delivered) {
 $topicName = sprintf("test_rdkafka_%s", uniqid());
 
 $consumer = new RdKafka\Consumer($conf);
-$consumer->addBrokers(TEST_KAFKA_BROKERS);
+$consumer->addBrokers(getenv('TEST_KAFKA_BROKERS'));
 
 $topic = $consumer->newTopic($topicName);
 $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
 
 while (true) {
     $msg = $topic->consume(0, 1000);
-    if ($msg && $msg->err) {
+    // librdkafka before 1.0 returns message with RD_KAFKA_RESP_ERR__PARTITION_EOF when reaching topic end.
+    if (!$msg || RD_KAFKA_RESP_ERR__PARTITION_EOF === $msg->err) {
         break;
     }
 }
 
 $topic->consumeStop(0);
 --EXPECT--
-
--- a/tests/conf_callbacks_integration.phpt
+++ b/tests/conf_callbacks_integration.phpt
@@ -2,38 +2,18 @@
 RdKafka\Conf
 --SKIPIF--
 <?php
-RD_KAFKA_VERSION >= 0x090000 || die("skip");
-file_exists(__DIR__."/test_env.php") || die("skip");
+RD_KAFKA_VERSION >= 0x090000 || die("skip librdkafka too old");
+(!isset($_ENV['TESTS_DONT_SKIP_RISKY']) || $_ENV['TESTS_DONT_SKIP_RISKY']) && die("skip Callbacks often fail and are skipped by default");
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $conf = new RdKafka\Conf();
 
-$topicConf = new RdKafka\TopicConf();
-$topicConf->set('auto.offset.reset', 'smallest');
-
-$conf->setDefaultTopicConf($topicConf);
-$conf->set('metadata.broker.list', TEST_KAFKA_BROKERS);
+$conf->set('auto.offset.reset', 'smallest');
+$conf->set('metadata.broker.list', getenv('TEST_KAFKA_BROKERS'));
 $conf->set('group.id', sprintf("test_rdkafka_group_%s", uniqid()));
-$conf->set('statistics.interval.ms', 10);
-
-$conf->setOffsetCommitCb(function ($consumer, $error, $topicPartitions) {
-    echo "Offset " . $topicPartitions[0]->getOffset() . " committed.\n";
-});
-
-$consumerLagFound = false;
-$conf->setStatsCb(function ($consumer, $json) use (&$consumerLagFound) {
-    if ($consumerLagFound) {
-        return;
-    }
-
-    // At some point there should be a consumer lag of 9
-    if (false !== strpos($json, 'consumer_lag":9')) {
-        $consumerLagFound = true;
-    }
-});
 
 $producer = new RdKafka\Producer($conf);
 
@@ -52,29 +32,44 @@ while ($producer->getOutQLen()) {
 // Make sure there is enough time for the stats_cb to pick up the consumer lag
 sleep(1);
 
+$conf = new RdKafka\Conf();
+
+$conf->set('auto.offset.reset', 'smallest');
+$conf->set('metadata.broker.list', getenv('TEST_KAFKA_BROKERS'));
+$conf->set('group.id', sprintf("test_rdkafka_group_%s", uniqid()));
+$conf->set('statistics.interval.ms', 10);
+
+$conf->setOffsetCommitCb(function ($consumer, $error, $topicPartitions) {
+    echo "Offset " . $topicPartitions[0]->getOffset() . " committed.\n";
+});
+
+$statsCbCalled = false;
+$conf->setStatsCb(function ($consumer, $json) use (&$statsCbCalled) {
+    if ($statsCbCalled) {
+        return;
+    }
+
+    $statsCbCalled = true;
+});
+
 $consumer = new RdKafka\KafkaConsumer($conf);
 $consumer->subscribe([$topicName]);
 
 while (true) {
-    $msg = $consumer->consume(60 * 1000);
+    $msg = $consumer->consume(15000);
 
-    if (!$msg) {
-        continue;
+    if (!$msg || RD_KAFKA_RESP_ERR__PARTITION_EOF === $msg->err) {
+        break;
     }
 
-    switch ($msg->err) {
-        case RD_KAFKA_RESP_ERR_NO_ERROR:
-            $consumer->commit($msg);
-
-            break;
-        case RD_KAFKA_RESP_ERR__PARTITION_EOF:
-            break 2;
-        default:
-            throw new Exception($msg->errstr());
+    if (RD_KAFKA_RESP_ERR_NO_ERROR !== $msg->err) {
+        throw new Exception($msg->errstr(), $msg->err);
     }
+
+    $consumer->commit($msg);
 }
 
-var_dump($consumerLagFound);
+var_dump($statsCbCalled);
 
 --EXPECT--
 Offset 1 committed.
--- a/tests/integration-tests-check.php
+++ b/tests/integration-tests-check.php
@@ -1,0 +1,11 @@
+<?php
+
+if (file_exists(__DIR__ . "/test_env.php")) {
+    include __DIR__ . '/test_env.php';
+}
+
+if (getenv('TEST_KAFKA_BROKERS')) {
+    return;
+}
+
+die('skip due to missing TEST_KAFKA_BROKERS environment & no test_env.php');
--- a/tests/message_headers.phpt
+++ b/tests/message_headers.phpt
@@ -2,12 +2,11 @@
 Message headers
 --SKIPIF--
 <?php
-RD_KAFKA_VERSION >= 0x000b04ff || die("skip");
-file_exists(__DIR__."/test_env.php") || die("skip");
+RD_KAFKA_VERSION >= 0x000b04ff || die("skip librdkafka too old");
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $delivered = 0;
 
@@ -25,7 +24,7 @@ $conf->setDrMsgCb(function ($producer, $msg) use (&$delivered) {
 
 $producer = new RdKafka\Producer($conf);
 
-if ($producer->addBrokers(TEST_KAFKA_BROKERS) < 1) {
+if ($producer->addBrokers(getenv('TEST_KAFKA_BROKERS')) < 1) {
     echo "Failed adding brokers\n";
     exit;
 }
@@ -62,7 +61,7 @@ while ($producer->getOutQLen()) {
 printf("%d messages delivered\n", $delivered);
 
 $consumer = new RdKafka\Consumer($conf);
-$consumer->addBrokers(TEST_KAFKA_BROKERS);
+$consumer->addBrokers(getenv('TEST_KAFKA_BROKERS'));
 
 $topic = $consumer->newTopic($topicName);
 $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
@@ -70,29 +69,25 @@ $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
 $messages = [];
 
 while (true) {
-    $msg = $topic->consume(0, 60*1000);
-    if (!$msg) {
-        continue;
+    $msg = $topic->consume(0, 1000);
+    if (!$msg || $msg->err === RD_KAFKA_RESP_ERR__PARTITION_EOF) {
+        break;
     }
-    switch ($msg->err) {
-        case RD_KAFKA_RESP_ERR_NO_ERROR:
-            $headersString = $msg->headers ?? [];
-            array_walk($headersString, function(&$value, $key) {
-                $value = "{$key}: {$value}";
-            });
-            if (empty($headersString)) {
-                $headersString = "none";
-            } else {
-                $headersString = implode(", ", $headersString);
-            }
-            printf("Got message: %s | Headers: %s\n", $msg->payload, $headersString);
-            break;
-        case RD_KAFKA_RESP_ERR__PARTITION_EOF:
-            echo "EOF\n";
-            break 2;
-        default:
-            throw new Exception($message->errstr());
+
+    if (RD_KAFKA_RESP_ERR_NO_ERROR !== $msg->err) {
+        throw new Exception($msg->errstr(), $msg->err);
     }
+
+    $headersString = isset($msg->headers) ? $msg->headers : [];
+    array_walk($headersString, function(&$value, $key) {
+        $value = "{$key}: {$value}";
+    });
+    if (empty($headersString)) {
+        $headersString = "none";
+    } else {
+        $headersString = implode(", ", $headersString);
+    }
+    printf("Got message: %s | Headers: %s\n", $msg->payload, $headersString);
 }
 --EXPECT--
 5 messages delivered
@@ -101,4 +96,3 @@ Got message: message 1 | Headers: key1: value1, key2: value2, key3: value3
 Got message: message 2 | Headers: none
 Got message: message 3 | Headers: none
 Got message: message 4 | Headers: none
-EOF
--- a/tests/produce_consume.phpt
+++ b/tests/produce_consume.phpt
@@ -2,17 +2,16 @@
 Produce, consume
 --SKIPIF--
 <?php
-file_exists(__DIR__."/test_env.php") || die("skip");
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $delivered = 0;
 
 $conf = new RdKafka\Conf();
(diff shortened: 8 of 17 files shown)


tests/conf_callbacks_integration.phpt
```diff
@@ -23,16 +23,13 @@ $conf->setOffsetCommitCb(function ($consumer, $error, $topicPartitions) {
...
-$conf->setStatsCb(function ($consumer, $json) use (&$consumerLagFound) {
-    if ($consumerLagFound) {
+$statsCbCalled = false;
+$conf->setStatsCb(function ($consumer, $json) use (&$statsCbCalled) {
+    if ($statsCbCalled) {
         return;
     }
 
-    // At some point there should be a consumer lag of 9
-    if (false !== strpos($json, 'consumer_lag":9')) {
```
Steveb-p, line 33: On `librdkafka` `1.0.0` consumer_lag property was never different than `-1` or `0` for me.

tests/message_headers.phpt
```diff
@@ -100,5 +97,4 @@ Got message: message 0 | Headers: key: value
 Got message: message 1 | Headers: key1: value1, key2: value2, key3: value3
 Got message: message 2 | Headers: none
 Got message: message 3 | Headers: none
-Got message: message 4 | Headers: none
-EOF
```
Steveb-p, line 104: This test also indicated memory leaks when `--enable-debug` php build was used.

```
5 messages delivered
Got message: message 0 | Headers: key: value
Got message: message 1 | Headers: key1: value1, key2: value2, key3: value3
Got message: message 2 | Headers: none
Got message: message 3 | Headers: none
Got message: message 4 | Headers: none
[Fri May 10 03:29:22 2019]  Script:  '/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php'
/home/steveb/CLionProjects/php-src/Zend/zend_string.h(133) :  Freeing 0x00007fce02464e00 (32 bytes), script=/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php
Last leak repeated 7 times
[Fri May 10 03:29:22 2019]  Script:  '/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php'
/home/steveb/CLionProjects/php-src/Zend/zend_string.h(133) :  Freeing 0x00007fce02464e40 (32 bytes), script=/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php
Last leak repeated 7 times
[Fri May 10 03:29:22 2019]  Script:  '/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php'
/home/steveb/CLionProjects/php-src/Zend/zend_hash.c(224) :  Freeing 0x00007fce02465c60 (56 bytes), script=/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php
Last leak repeated 6 times
[Fri May 10 03:29:22 2019]  Script:  '/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php'
/home/steveb/CLionProjects/php-src/Zend/zend_hash.c(132) :  Freeing 0x00007fce02470080 (320 bytes), script=/home/steveb/CLionProjects/php-rdkafka/tests/message_headers.php
Last leak repeated 3 times
=== Total 27 memory leaks detected ===
```

Steveb-p:
@nick-zh I've rebased the branch and ran tests again - there is particularly one test failing: `tests/conf_callbacks_integration.phpt`. It throws `Exception` with code -185, which gets us `Local: Timed Out` after calling `$msg->errstr()`.

I've added Dockerfile with build and docker-compose.yaml to allow us to test locally in the same environment. I'll remove them once we're ok with results.

I'm running tests using `docker-compose run --rm php make test`.

Steveb-p:
We discussed this with @nick-zh in conversation.

Currently all tests integration tests are green apart from the one with statistic callback enabled - this one causes local librdkafka error (`-185`, which translated to `Local: Timed Out` on librdkafka > 1.0.0). This does not prevent user from properly reading statistic information or producing/consuming messages, but may indicate a problem somewhere.

I'll be working on introducing integration tests into Travis pipeline, so they will be executed as well.

Steveb-p:
Updated the PR with changes to CI pipeline:

- Split build and test parts of travis script
- Moved both building & testing into docker containers. At first I tried running Kafka/Zookeeper containers and run into an issue with one of the tests failing. Tried the same in containers and it passed. However, after a couple more runs it became apparent that the test is unstable and passes in roughly 50% of cases, so my initial assumption that host -> docker connectivity was at fault was wrong. I can revert to testing from Travis host environment.
- Tests now primarily use environment variables to decide if integration tests should run. You can still use test_env.php file, but I'd say we should eventually drop it?
- Added librdkafka 1.2.2 to build matrix and made librdkafka repository configurable if needed (with sensible defaults)

I've decided to limit test matrix temporarily to just one job to speed up the process of adjusting tests, and will re-enable others once we're happy with them.

Thoughts @arnaud-lb @nick-zh? What can be the cause of this single test failing? There are other Consumer tests before it and they're not affected by described instability.

EDIT: You can run tests locally with docker using:
```bash
docker-compose up -d kafka
# build accepts arguments, so you can switch librdkafka / PHP version
docker-compose build
docker-compose run --rm php
# or simply "docker-compose up php"
```

nick-zh:
@Steveb-p i will have a look over the weekend, sry ultra busy this week, so i probably won't find time earlier

.travis.yml
```diff
@@ -1,25 +1,56 @@
...
         packages:
             - valgrind
-php:
-    - 5.6
-    - 7.0
-    - 7.1
-    - 7.2
-    - 7.3
-    - 7.4snapshot
-    - nightly
```
nick-zh, line 16: would be nice to still have nightly

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+FROM php:$PHP_IMAGE_VERSION
+
+ARG LIBRDKAFKA_VERSION='v1.2.1'
+ARG LIBRDKAFKA_REPOSITORY_URL='https://github.com/edenhill/librdkafka.git'
+
+RUN docker-php-source extract \
+    && apt-get update \
+    && apt-get install -y \
+    $PHPIZE_DEPS \
```
nick-zh: maybe i overlooked it, but i was unalbe to finde the definition of `$PHPIZE_DEPS`

.travis.yml
```diff
@@ -1,25 +1,56 @@
...
-    - LIBRDKAFKA_VERSION=v1.1.0
-    - LIBRDKAFKA_VERSION=master
-script: ./.travis/build.sh
+    global:
+        # TEST_KAFKA_BROKER* environment variables are overwritten in docker-compose.yaml
+        - TEST_KAFKA_BROKERS=kafka:9093
+        - TEST_KAFKA_BROKER_VERSION=2.3
+        - LIBRDKAFKA_REPOSITORY_URL=https://github.com/edenhill/librdkafka.git
+    jobs:
+        - LIBRDKAFKA_VERSION=0.11.x PHP_IMAGE_VERSION=5.6
```
nick-zh:
```suggestion
        - LIBRDKAFKA_VERSION=0.11.x PHP_IMAGE_VERSION=5.6-alpine
```
i think it would be nice to actually use the alpine images, since they are about a 5th of the debian images

docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
+version: '3.7'
+
+services:
+  php:
+    build:
+      context: .
```
nick-zh:
```suggestion
      context: .
      args:
        LIBRDKAFKA_VERSION: ${LIBRDKAFKA_VERSION}
        PHP_IMAGE_VERSION: ${PHP_IMAGE_VERSION}
```

.travis.yml
```diff
@@ -1,25 +1,56 @@
...
+        - LIBRDKAFKA_VERSION=master PHP_IMAGE_VERSION=7.0
+        - LIBRDKAFKA_VERSION=master PHP_IMAGE_VERSION=7.1
+        - LIBRDKAFKA_VERSION=master PHP_IMAGE_VERSION=7.2
+        - LIBRDKAFKA_VERSION=master PHP_IMAGE_VERSION=7.3
+        - LIBRDKAFKA_VERSION=master PHP_IMAGE_VERSION=7.4-rc
+script: 
+    ### Run Kafka in docker. Starting it up early also makes sure it's online after build
+    - docker-compose up -d kafka zookeeper
+    - docker-compose build --build-arg LIBRDKAFKA_VERSION="${LIBRDKAFKA_VERSION}" --build-arg PHP_IMAGE_VERSION="${PHP_IMAGE_VERSION}" php
+    - docker-compose run --rm php
```
nick-zh: intstead of the three commands you have now, i would probably just do a:
```
docker-compose up --build -d
```
with the changes in `docker-compose.yml` we shouldn't be needing the build args here.
If we see we have a timing issue, we can add a little loop here (or in a script) which waits for kafka to be available on its port. I think starting it like this, might still not guarantee that you don't run into timing issues.

Dockerfile
```diff
@@ -0,0 +1,30 @@
+ARG PHP_IMAGE_VERSION='latest'
+
+FROM php:$PHP_IMAGE_VERSION
+
+ARG LIBRDKAFKA_VERSION='v1.2.1'
```
nick-zh:
```suggestion
ARG LIBRDKAFKA_VERSION='v1.2.2'
```

docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
...
+    depends_on:
+      - zookeeper
+      - kafka
+    environment:
+      TEST_KAFKA_BROKERS: "${TEST_KAFKA_BROKERS:-kafka:9093}"
+      TEST_KAFKA_BROKER_VERSION: "${TEST_KAFKA_BROKER_VERSION:-2.3}"
+    volumes:
+      - ./tests:/build/tests
```
nick-zh: since we copy the whole source into the container, do you think this is still needed?

CONTRIBUTING.md
```diff
@@ -10,4 +10,35 @@ Pull requests should be made against the master branch, which supports both PHP
...
+
+```
+Substituting `<KAFKA_HOSTNAME>`, `<KAFKA_PORT>` and `<KAFKA_VERSION>` with values proper for your kafka instance.
+
+2. Running tests in docker.
+You can run those by executing
+```shell script
+docker-compsoe up -d kafka
+docker-compose build
+docker-compose run --rm php
```
nick-zh: if we change the script for travis, adjust here accordingly as well ofc 
very nice addition, i like it very much :heart:

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ARG LIBRDKAFKA_REPOSITORY_URL='https://github.com/edenhill/librdkafka.git'
+
+RUN docker-php-source extract \
+    && apt-get update \
```
nick-zh: when switching to alpine, to install stuff, you can simply use:
```
apk add $PHPIZE_DEPS bash git valgrind
```
i think the source extract is not needed (at least i don't rembember using it for alpine)

tests/allow_null_payload_and_key.phpt
```diff
@@ -2,11 +2,16 @@
 Allow null payload
 --SKIPIF--
 <?php
-file_exists(__DIR__."/test_env.php") || die("skip");
+getenv('TEST_KAFKA_BROKERS') !== false || file_exists(__DIR__."/test_env.php") || die("skip");
 --FILE--
 <?php
 
-require __DIR__."/test_env.php";
+if (getenv('TEST_KAFKA_BROKERS') === false) {
```
nick-zh: instead of changing this, what i would do is, i would add another file `tests/test_env.php.template`
with the following content:
```
define('TEST_KAFKA_BROKERS', ${TEST_KAFKA_BROKERS});
define('TEST_KAFKA_BROKER_VERSION', ${TEST_KAFKA_BROKER_VERSION});
```
make sure that we have those envs during the build of our docker image.
Make sure you have also installed `gettext` (alpine).
Then what you can do is after you have copied the source in the docker image:
``` 
RUN envsubst < "tests/test_env.php.template" > "tests/test_env.php"
```
with this, we can probably even get rid of the file_exists check

nick-zh (changes requested):
i did a review of the docker setup and left some notes,
i will be checking the actuall test changes later today :v:

docker-compose.yaml
Steveb-p: I used it to get test results back to host filesystem. Since .phpt's usually don't output useful information (or at least I'm not familiar with those types of tests to set them up properly) I used it to look over expected/actual/diff.

.travis.yml
Steveb-p: I'm going to revert changes to travis CI in this regard. Building PHP in "main travis thread" should be sufficient, and I only wanted to use Kafka image to ease starting of it's service. 

The only reason I actually removed PHP from Travis was to check if those builds behave the same way in docker containers (with tests failing on particular occasions, which they do :disappointed: )

.travis.yml
Steveb-p: I didn't want to use alpine, because in general use case users build phprdkafka in context of debian/ubuntu servers and I wanted to "test" relevant installation instructions this way.

IMO we could provide instructions how to build on alpine, if any steps differ from "normal" linux OS.

tests/allow_null_payload_and_key.phpt
Steveb-p: I've added `tests/integration-tests-check.php` that should do environment variables handling/check and should be loaded before each integration test. I've also switched from using constants to environment variables everywhere.

Dockerfile
Steveb-p: I'm not sure but I think that without php sources valgrind will report false positives in memory leaks? But I'm not sure, I was having issues with valgrind anyway :confused:

CONTRIBUTING.md
Steveb-p: While I like it as well, as I said, there are some caveats that occur with docker-based Kafka :confused: Until they are sorted they will cause tests to fail in such environment.

docker-compose.yaml
Steveb-p: Adjusted accordingly.

Steveb-p:
I've managed to get those integration tests to work (finally). 

I had to scrap docker completely as it seems that there are some issues stemming from it. I've went back to keeping everything in host OS and I have found that **we have probably broke the `producev` method** at `ed6c5e9837cbcaac873b31701335edcf00590bbe`. I've rebased the branch just before that commit and all integration tests pass. As soon as those changes are added, calling `producev` results in early termination by `SIGTERM` signal.

nick-zh:
@Steveb-p ok, good to know, i am assuming you will revert the travis / docker changes then? I think i will have a look at a later point to add docker test support.
About the sigterm, i will have a look, i think i tested the use of producev, but i will have another look.
If there is some additional trace / log you could provide, that would be also great.

Steveb-p:
> ok, good to know, i am assuming you will revert the travis / docker changes then?

Already done that, only left `Dockerfile` and `docker-compose.yaml` atm. `.travis.yaml` has been changed to work with host OS again only.

> About the sigterm, i will have a look, i think i tested the use of producev, but i will have another look.
> If there is some additional trace / log you could provide, that would be also great.

I'm assuming you've got that yourself since you've got #318. I've cherry-picked the commit and we'll see how the tests go.

tests/bug115.phpt
```diff
@@ -28,14 +27,14 @@ $conf->setDrMsgCb(function ($producer, $msg) use (&$delivered) {
...
-$consumer->addBrokers(TEST_KAFKA_BROKERS);
+$consumer->addBrokers(getenv('TEST_KAFKA_BROKERS'));
 
 $topic = $consumer->newTopic($topicName);
 $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
 
 while (true) {
     $msg = $topic->consume(0, 1000);
-    if ($msg && $msg->err) {
+    if (!$msg) {
```
nick-zh:
```suggestion
    if (null === $msg || $msg->err === RD_KAFKA_RESP_ERR__PARTITION_EOF) {
```

tests/produce_consume.phpt
```diff
@@ -52,27 +51,27 @@ while ($producer->getOutQLen()) {
...
 
 $topic = $consumer->newTopic($topicName);
 $topic->consumeStart(0, RD_KAFKA_OFFSET_BEGINNING);
 
 $messages = [];
 
 while (true) {
-    $msg = $topic->consume(0, 60*1000);
+    $msg = $topic->consume(0, 1000);
     if (!$msg) {
```
nick-zh:
```suggestion
    // EOF reached
    if (null === $msg || $msg->err === RD_KAFKA_RESP_ERR__PARTITION_EOF) {
```
This way we have it in one place and can remove the switch below, for me it would be less confusing if i was new to this

tests/produce_consume_queue.phpt
```diff
@@ -69,20 +68,20 @@ array_walk($topicNames, function ($topicName) use ($consumer, $queue) {
 $messages = [];
 $eof = 0;
 
-while ($eof < 2) {
-    $msg = $queue->consume(60*1000);
+while (true) {
+    $msg = $queue->consume(1000);
     if (!$msg) {
```
nick-zh, line 75: same here with the eof as above :smiley:

tests/produce_consume_queue.phpt
```diff
@@ -69,20 +68,20 @@ array_walk($topicNames, function ($topicName) use ($consumer, $queue) {
 $messages = [];
 $eof = 0;
```
nick-zh: if this count is obsolete, can also be removed here

tests/integration-tests-check.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
nick-zh, line 1: i would actually remove this file and put the checks back as they were and just do the adjustments in the sample

tests/bug115.phpt
```diff
@@ -2,17 +2,16 @@
...
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $delivered = 0;
 
 $conf = new RdKafka\Conf();
 if (RD_KAFKA_VERSION >= 0x090000) {
-    $conf->set('broker.version.fallback', TEST_KAFKA_BROKER_VERSION);
+    $conf->set('broker.version.fallback', getenv('TEST_KAFKA_BROKER_VERSION'));
```
nick-zh, line 14: I actually think we can get rid of the version everywhere, the default should works for us and we are not using multiple versions of Kafka.

tests/test_env.php.sample
```diff
@@ -0,0 +1,4 @@
+<?php
+
+putenv('TEST_KAFKA_BROKERS=localhost:9092');
+putenv('TEST_KAFKA_BROKER_VERSION=2.3');
```
nick-zh:
```suggestion
```

tests/test_env.php.sample
```diff
@@ -0,0 +1,4 @@
+<?php
+
+putenv('TEST_KAFKA_BROKERS=localhost:9092');
```
nick-zh:
```suggestion
if(false === getenv('TEST_KAFKA_BROKERS')) {
    putenv('TEST_KAFKA_BROKERS=localhost:9092');
}
```

.travis.yml
```diff
@@ -12,14 +12,27 @@ php:
...
         - php: nightly
 env:
-    - LIBRDKAFKA_VERSION=0.11.x
-    - LIBRDKAFKA_VERSION=v1.0.1
-    - LIBRDKAFKA_VERSION=v1.1.0
-    - LIBRDKAFKA_VERSION=master
-script: ./.travis/build.sh
+    global:
+        - TEST_KAFKA_BROKERS=localhost:9092
+        - TEST_KAFKA_BROKER_VERSION=2.3
```
nick-zh, line 23:
```suggestion
```

.travis.yml
```diff
@@ -12,14 +12,27 @@ php:
...
+        - LIBRDKAFKA_REPOSITORY_URL=https://github.com/edenhill/librdkafka.git
+        - LIBRDKAFKA_VERSION=v1.2.2
```
nick-zh, line 25:
```suggestion
```
don't think we need it as globabl, but not sure, let me know

nick-zh (changes requested):

tests/test_env.php.sample
Steveb-p: You want this environment variable to be removed?

tests/bug115.phpt
nick-zh: nah i was wrong, i checked, the default was changed after `librdkafka:0.11.6` and since we still support that, we need it for now.

.travis.yml
Steveb-p: Travis behaves funny if it's missing (afair it doesn't apply matrix properly) and I've seen it in Doctrine ODM as well, so it's not "just us".

tests/integration-tests-check.php
Steveb-p: I'm leaning towards leaving it to allow local testing. This ensures that you can use either environment variables or `test_env.php` file to allow integration tests - and you can opt-out if you want, if you're not running local kafka instance (not to mention the amount of time those tests take, which is far more than "unit" ones).

.travis.yml
Steveb-p: see https://github.com/doctrine/mongodb-odm/blob/master/.travis.yml#L17

Steveb-p:
All commits squashed. I'd say it's as good as it's gonna get, and we can still work on this after it's merged (integration tests missing really hurt us with `producev` method :disappointed: ).

We can add more tests further on.

Care for re-review @nick-zh? Any comments @arnaud-lb?

arnaud-lb:
Great work @Steveb-p ! 

I see that you added a `Dockerfile` / `docker-compose.yml`, where is this used ?

Just a note on `RD_KAFKA_RESP_ERR__PARTITION_EOF`: Messages with this error code are returned only when `enable.partition.eof` is `true` in the conf.

Steveb-p:
> I see that you added a `Dockerfile` / `docker-compose.yml`, where is this used ?

It's a leftover from attempting to simplify local testing by providing an "automated" and consistent build and testing tool. I backed down from it since callback integration tests failed with a high rate, only to see them fail as well - albeit with roughly 50/50 change - on normal, host-based testing.

That's why this test is currently skipped, so integration tests can be enabled to prevent issues like the one with `producev` and we can work on introducing more test cases / organize them better.

All in all, I'll remove those files and put them back with instructions in a separate PR.

> Just a note on `RD_KAFKA_RESP_ERR__PARTITION_EOF`: Messages with this error code are returned only when `enable.partition.eof` is `true` in the conf.

Practically it's only checked for tests for librdkafka 0.11.6, since the defaults changed afaik? We can switch to setting `enable.partition.eof` always instead, but I didn't want to change tests too much.

nick-zh:
@Steveb-p `enable.partition.eof` is `true` for anything before `librdkafka:1.0.0`

tests/produce_consume_queue.phpt
```diff
@@ -67,23 +66,19 @@ array_walk($topicNames, function ($topicName) use ($consumer, $queue) {
...
-            break;
-        default:
-            throw new Exception($message->errstr());
+
+    if (RD_KAFKA_RESP_ERR_NO_ERROR !== $msg->err) {
+        throw new Exception($msg->errstr(), $msg->err);
     }
+
+    $messages[] = sprintf("Got message: %s from %s", $msg->payload, $msg->topic_name);
 }
```
nick-zh:
```
while (true) {
    $msg = $queue->consume(15000);
    // librdkafka before 1.0 returns message with RD_KAFKA_RESP_ERR__PARTITION_EOF when reaching topic end.
    if (!$msg) {
        continue;
    }

    if ($msg->err === RD_KAFKA_RESP_ERR__PARTITION_EOF) {
        break;
    } elseif (RD_KAFKA_RESP_ERR_NO_ERROR !== $msg->err) {
        throw new Exception($msg->errstr(), $msg->err);
    }
    $messages[] = sprintf("Got message: %s from %s", $msg->payload, $msg->topic_name);
}
```

tests/produce_consume_queue.phpt
```diff
@@ -2,17 +2,16 @@
...
+require __DIR__ . '/integration-tests-check.php';
 --FILE--
 <?php
-
-require __DIR__."/test_env.php";
+require __DIR__ . '/integration-tests-check.php';
 
 $delivered = 0;
 
 $conf = new RdKafka\Conf();
```
nick-zh, line 12:
```suggestion
$conf = new RdKafka\Conf();
$conf->set('enable.partition.eof', 'true');
```

nick-zh (changes requested):

nick-zh:
@arnaud-lb LGTM, please let us know if we need more, otherwise should be good to merge :tada: 
Thx a lot @Steveb-p , with docker in place, makes it a bit easier for me to run this locally :heart:

nick-zh (approved):

arnaud-lb:
Thanks @Steveb-p !

PR #228, opened 2019-06-04
Support sending timestamp (epoch ms) to producev

lkm:
This will add support for setting timestamp for messages produced using the `producev` method. The format is epoch in milliseconds

### DIFF
--- a/topic.c
+++ b/topic.c
@@ -402,8 +402,8 @@ PHP_METHOD(RdKafka__ProducerTopic, produce)
 /* }}} */
 
 #ifdef HAVE_RD_KAFKA_MESSAGE_HEADERS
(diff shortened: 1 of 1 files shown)


topic.c
```diff
@@ -429,8 +430,10 @@ PHP_METHOD(RdKafka__ProducerTopic, producev)
     char *header_key;
     zeval *header_value;
     rd_kafka_headers_t *headers;
+    long timestamp_ms;
+    zend_bool timestamp_ms_is_null;
```
arnaud-lb: I think that you need to initialize these two variables here. Optional parameters can be left uninitialized by `zend_parse_parameters`.

lkm: Thanks @arnaud-lb. That's been fixed

arnaud-lb:
Thanks !

PR #236, opened 2019-07-09
Add PHP 7.3 and librdkafka 1.0.x into build matrix

renatomefi:


nick-zh:
@renatomefi would probably make sense to add librdkafka 1.1.x as well

renatomefi:
Hey @nick-zh, 1.1 branch isn't available yet since it's master, they'll probably add it when 1.2 is out!

nick-zh:
Hey @renatomefi ah yeah sry, i see now
Seems 7.3 builds are failing :cry:

arnaud-lb:
Thanks @renatomefi. PHP 7.3 added some asm for string comparison that intentionally reads uninitialized memory (see https://bugs.php.net/bug.php?id=76445), causing valgrind to report some issues. I'm adding 7.3 in https://github.com/arnaud-lb/php-rdkafka/pull/249.

Issue #130, opened 2017-09-13
Allow users to search offsets by timestamp on partitions


chrisLeeTW:
As of 0.10.0.1 the java library lets users seek to offsets based on timestamp.

https://kafka.apache.org/documentation#upgrade_1010_notable

And also offsets_for_times() support was added to librdkafka.

nick-zh:
@arnaud-lb any update / plans on this? Having support for https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2695 would be really great.
Sadly my C knowledge is very very rusty, so it would take me quite a while to contribute this.
Cheers,
Nick

nick-zh:
@arnaud-lb i think we can close this after doc is merged, thanks in advance :+1:

PR #238, opened 2019-07-12
Add offset for times

nick-zh:
https://github.com/arnaud-lb/php-rdkafka/issues/130

### DIFF
--- a/rdkafka.c
+++ b/rdkafka.c
@@ -589,6 +589,40 @@ PHP_METHOD(RdKafka__Kafka, queryWatermarkOffsets)
 /* }}} */
 #endif /* HAVE_RD_KAFKA_QUERY_WATERMARK_OFFSETS */
 
+#ifdef HAVE_RD_KAFKA_OFFSETS_FOR_TIMES
+/* {{{ proto void RdKafka\Kafka::offsetsForTimes(array $topicPartitions, int $timeout_ms)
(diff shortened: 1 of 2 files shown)


nick-zh:
@arnaud-lb if you find the time, i would appreciate if you can have a look and check if i did everything properly, since this is my first contribution. Thx in advance man :smile:

rdkafka.c
```diff
@@ -589,6 +589,48 @@ PHP_METHOD(RdKafka__Kafka, queryWatermarkOffsets)
...
+    topicPartitions = array_arg_to_kafka_topic_partition_list(1, htopars TSRMLS_CC);
+    if (!topicPartitions) {
+        return;
+    }
+
+    err = rd_kafka_offsets_for_times(intern->rk, topicPartitions, timeout_ms);
+
+    if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
+        zend_throw_exception(ce_kafka_exception, rd_kafka_err2str(err), err TSRMLS_CC);
+        return;
```
arnaud-lb, line 626: We might need to destroy topicPartitions before return here

nick-zh: i added a destroy before throwing the exception, thanks for the feedback

arnaud-lb:
Thanks !

PR #245, opened 2019-07-25
add block msg flag

nick-zh:
I am not sure if this is what you wanted for https://github.com/arnaud-lb/php-rdkafka/issues/79
If the intent was something else, please let me know


nick-zh:
I see that this flag has been added in `librdkafka:0.9.2` this is way the earlier tests (0.8.x) are failing.
What do you propose @arnaud-lb ?

arnaud-lb:
By default, produce() is non blocking: The message is added in a queue, that is handled in the background. This flag optionally makes produce() wait until the message is actually sent.

We should let the user optionally pass this flags in the `$msgflags` parameter of `produce()`. We could do this by exposing `RD_KAFKA_MSG_F_BLOCK` as a PHP constant.

For librdkafka 0.8.x: We can use `#ifdef` in the code to check if `RD_KAFKA_MSG_F_BLOCK` is defined.

Steveb-p:
> For librdkafka 0.8.x: We can use `#ifdef` in the code to check if `RD_KAFKA_MSG_F_BLOCK` is defined.

There will come a time that maintaining that old version of librdkafka will simply be an additional maintenance with little to no benefit. 0.8 is already quite old and considering the amount of work put into 1.0 conflicts are bound to happen.

Maybe consider droping support to anything that is < 0.10?

nick-zh:
@arnaud-lb thx for the input, i'll look into it
Regarding @Steveb-p's comment, what are your thoughts / policies for dropping old versions?
`0.8` dates back to 2013
`0.9` to late 2015
`0.11` from 2017, but even with the `0.11.x` releases there have been a lot of critical fixes, the latest one in this bunch being `0.11.6` which i actually used for quite some time and seemed very stable.
I think dropping support for anything before `0.11` would be a reasonable thing to do, to also avoid hunting issues that are lib related.
@arnaud-lb please let us know what you think about this :+1:

nick-zh:
I made the constant `RD_KAFKA_MSG_F_BLOCK` available and i am explicitly checking for either `0`  or for `RD_KAFKA_MSG_F_BLOCK` since i think, this are the only two options that we want to provide. You alread add `RD_KAFKA_MSG_F_COPY` per default, so `RD_KAFKA_MSG_F_FREE` does not seem to make sense anyway @arnaud-lb

config.m4
```diff
@@ -88,6 +88,12 @@ if test "$PHP_RDKAFKA" != "no"; then
     AC_MSG_WARN([no rd_kafka_offsets_for_times, offsetsForTimes will not be available])
   ])
 
+  AC_CHECK_DECLS([RD_KAFKA_MSG_F_BLOCK],[
+    AC_DEFINE(HAVE_RD_KAFKA_MSG_F_BLOCK,1,[ ])
+  ],[
+    AC_MSG_WARN([no RD_KAFKA_MSG_F_BLOCK, flag for blocking produce, on full queue, will not be available])
+  ],[[#include <librdkafka/rdkafka.h>]])
```
arnaud-lb: Since RD_KAFKA_MSG_F_BLOCK is a constant, we could just use `#ifdef RD_KAFKA_MSG_F_BLOCK` in the code, and we would not need to add anything in `config.m4`. What do you think ?

rdkafka.c
```diff
@@ -805,6 +805,9 @@ PHP_MINIT_FUNCTION(rdkafka)
     COPY_CONSTANT(RD_KAFKA_OFFSET_STORED);
     COPY_CONSTANT(RD_KAFKA_PARTITION_UA);
     COPY_CONSTANT(RD_KAFKA_PRODUCER);
+#ifdef HAVE_RD_KAFKA_MSG_F_BLOCK
+    REGISTER_LONG_CONSTANT("RD_KAFKA_MSG_F_BLOCK", RD_KAFKA_MSG_F_BLOCK, CONST_CS | CONST_PERSISTENT);
+#endif
```
arnaud-lb: Would `COPY_CONSTANT()` work here ?

topic.c
```diff
@@ -384,7 +384,7 @@ PHP_METHOD(RdKafka__ProducerTopic, produce)
         return;
     }
 
-    if (msgflags != 0) {
+    if (msgflags != 0 && msgflags != RD_KAFKA_MSG_F_BLOCK) {
```
nick-zh, line 387: @arnaud-lb  do i need to switch this check to it's int value, since if an old rdkafka version is used, this is undefined?
```suggestion
    if (msgflags != 0 && msgflags != 0x4) {
```

topic.c
```diff
@@ -442,7 +442,7 @@ PHP_METHOD(RdKafka__ProducerTopic, producev)
...
```
nick-zh, line 445:
```suggestion
    if (msgflags != 0 && msgflags != 0x4) {
```
same here

arnaud-lb:
@nick-zh @Steveb-p you are right, let's drop support of versions prior to 0.11. They are being packaged only in very old distro versions.

I will remove 0.8 and 0.9 from the build matrix.

nick-zh:
@arnaud-lb i am adjusting my PR accordingly, this will mean, that i don't need to check with ifdef anymore, since `RD_KAFKA_MSG_F_BLOCK` was added prior to 0.11
**Edit:** I updated the readme and the build matrix accordingly

arnaud-lb:
Merged, thanks !

Issue #233, opened 2019-07-01
how can I set the logger using rd_kafka_conf_set_log_cb


kirachen1991:
* PHP version:7.3.6
 * librdkafka version:1.1.0
 * php-rdkafka version:3.1.0

It seems like that I can set logger by using the function rd_kafka_conf_set_log_cb

But I can not find the way to do

I read the document, that function RdKafka\Conf::set can set rd_kafka_conf_set_log_cb, but
I don't know how to set……

nick-zh:
@kirachen1991 from what i see in the doc and the code, it seems this is not implemented in the lib yet, but if you have some C knowledge, here are the other callbacks, probably it would be not so hard to create a PR:
https://github.com/arnaud-lb/php-rdkafka/blob/master/conf.c
Relates to: https://github.com/arnaud-lb/php-rdkafka/issues/127

Steveb-p:
Awesome work @nick-zh :) This issue might be closed and/or a documentation one should be opened?

nick-zh:
@Steveb-p thank you, actually i already have a doc pr open for this: https://github.com/arnaud-lb/php-rdkafka-doc/pull/26
think this can be closed

PR #253, opened 2019-07-27
add log-cb php function

nick-zh:
#233

### DIFF
--- a/rdkafka.c
+++ b/rdkafka.c
@@ -147,7 +147,7 @@ static void kafka_init(zval *this_ptr, rd_kafka_type_t type, zval *zconf TSRMLS_
         return;
     }
 
-    if(intern->cbs.log) {
+    if (intern->cbs.log) {
         rd_kafka_set_log_queue(rk, NULL);
     }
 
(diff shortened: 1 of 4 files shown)


rdkafka.c
```diff
@@ -147,6 +147,10 @@ static void kafka_init(zval *this_ptr, rd_kafka_type_t type, zval *zconf TSRMLS_
         return;
     }
 
+    if (intern->cbs.log) {
+        rd_kafka_set_log_queue(rk, NULL);
```
arnaud-lb, line 156: Should we also set `log.queue=true` in the conf, before callling rd_kafka_new ? https://github.com/edenhill/librdkafka/blob/4fb7c9924c14f44c4f1490b80b8bd714098dd90b/src/rdkafka.h#L2923

nick-zh: @arnaud-lb yeah good input, i added to the doc that this needs to be true, but you are absolutely right, we can just set it actually, i will adjust this accordingly

nick-zh: @arnaud-lb done, let me know if the change is ok like that, sry for the style stuff :/ is there like a linter where i can check next time? (sry havent written c in some time)

arnaud-lb: Thanks :) I don't know C linters, but we should probably add one in the build at some point.

Issue #239, opened 2019-07-18
Consumer Batch message read


harshadvichare:
* PHP version: 7.0
 * librdkafka version: librdkafka-0.11.6
 * php-rdkafka version: php-rdkafka-3.0.5

How can we consume multiple message or Batch message read using rdKafka.
Also how to send  message with enable.idempotence = true

nick-zh:
@harshadvichare consuming of batches is not supported in the extension as for now
Regarding idempotence from what i understand from the release notes from librdkafka just setting that setting, should actually be enought during producing

harshadvichare:
@nick-zh Thanks for update 
Do you have any plan to support  for reading batch message in consumer.

nick-zh:
@harshadvichare is there a specific reason you prefer using batch consuming?

harshadvichare:
@nick-zh 
For the performance and bulk message consumption we need batch consuming

nick-zh:
@harshadvichare thx for getting back. If your sole concern is performance, it might be worth exposing the consume callback, in the lib there is the following statement:
```
`rd_kafka_consume_callback()` provides higher throughput performance
 than both `rd_kafka_consume()` and `rd_kafka_consume_batch()`.
```
I created a separate issue for this https://github.com/arnaud-lb/php-rdkafka/issues/242
If you need bulks, then of course batch makes sense as well.
I'll try to check on the weekend if i find time to expose this in the extension

@arnaud-lb could you please add the approriate labels here, thx :)

harshadvichare:
Thanks .

Is this batch reading functionality  in consumer is same as max.poll.records supported in kafka.
Current version of librdkafka not support max.poll.records. 
So can you please update what should be difference in both cases

nick-zh:
`rd_kafka_consume_batch` has an argument `rkmessages_size` which, at least in my understanding, returns the amount of messages you requested. I see there is an issue with discussions about `max.poll.records` but i think, this goes in a different direction, i will check and follow up here.

nick-zh:
@harshadvichare it is as suspected. There is the setting `max.poll.records` that controlls how many records will be returned from a `poll` and it seems `librdkafka` does not support it. So we'll have to set it through `rkmessages_size`

mcarbonneaux:
there an equivalent on high level consumer ?

nick-zh:
@mcarbonneaux there is not, no

mcarbonneaux:
While be nice to have!

[...]

PR #256, opened 2019-07-31
add batch consume

nick-zh:
#239

### DIFF
--- a/message.c
+++ b/message.c
@@ -85,6 +85,23 @@ void kafka_message_new(zval *return_value, const rd_kafka_message_t *message TSR
 #endif
 }
 
+void kafka_message_list_to_array(zval *return_value, rd_kafka_message_t *messages, long size TSRMLS_DC) /* {{{ */
+{
+    rd_kafka_message_t *msg;
+    zeval zmsg;
+    int i;
+
+    array_init_size(return_value, size);
+
+    for (i = 0; i < size; i++) {
+        msg = &messages[i];
+        MAKE_STD_ZEVAL(zmsg);
+        object_init_ex(P_ZEVAL(zmsg), ce_kafka_message);
+        kafka_message_new(P_ZEVAL(zmsg), msg TSRMLS_CC);
+        add_next_index_zval(return_value, P_ZEVAL(zmsg));
+    }
+} /* }}} */
+
 /* {{{ proto string RdKafka\Message::errstr()
  *  Returns the error string for an errored KrKafka\Message or NULL if there was no error.
  */
--- a/message.c
+++ b/message.c
@@ -85,7 +85,7 @@ void kafka_message_new(zval *return_value, const rd_kafka_message_t *message TSR
 #endif
 }
 
-void kafka_message_list_to_array(zval *return_value, rd_kafka_message_t *messages, long size TSRMLS_DC) /* {{{ */
+void kafka_message_list_to_array(zval *return_value, rd_kafka_message_t **messages, long size TSRMLS_DC) /* {{{ */
 {
     rd_kafka_message_t *msg;
     zeval zmsg;
@@ -94,7 +94,7 @@ void kafka_message_list_to_array(zval *return_value, rd_kafka_message_t *message
     array_init_size(return_value, size);
 
     for (i = 0; i < size; i++) {
-        msg = &messages[i];
+        msg = messages[i];
         MAKE_STD_ZEVAL(zmsg);
         object_init_ex(P_ZEVAL(zmsg), ce_kafka_message);
         kafka_message_new(P_ZEVAL(zmsg), msg TSRMLS_CC);
--- a/message.c
+++ b/message.c
@@ -96,7 +96,6 @@ void kafka_message_list_to_array(zval *return_value, rd_kafka_message_t **messag
     for (i = 0; i < size; i++) {
         msg = messages[i];
         MAKE_STD_ZEVAL(zmsg);
-        object_init_ex(P_ZEVAL(zmsg), ce_kafka_message);
         kafka_message_new(P_ZEVAL(zmsg), msg TSRMLS_CC);
         add_next_index_zval(return_value, P_ZEVAL(zmsg));
     }
--- a/topic.c
+++ b/topic.c
@@ -349,8 +349,7 @@ PHP_METHOD(RdKafka__ConsumerTopic, consumeBatch)
 
     if (result > 0) {
         kafka_message_list_to_array(return_value, rkmessages, batch_size TSRMLS_CC);
-        for (i = 0; i < batch_size; ++i)
-        {
+        for (i = 0; i < batch_size; ++i) {
             rd_kafka_message_destroy(rkmessages[i]);
         }
     }
--- a/topic.c
+++ b/topic.c
@@ -333,22 +333,23 @@ PHP_METHOD(RdKafka__ConsumerTopic, consumeBatch)
         return;
     }
 
-    rkmessages = malloc(sizeof(*rkmessages) * batch_size);
-
     intern = get_kafka_topic_object(getThis() TSRMLS_CC);
     if (!intern) {
         return;
     }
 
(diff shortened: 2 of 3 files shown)


message.c
```diff
@@ -85,6 +85,23 @@ void kafka_message_new(zval *return_value, const rd_kafka_message_t *message TSR
...
+    rd_kafka_message_t *msg;
+    zeval zmsg;
+    int i;
+
+    array_init_size(return_value, size);
+
+    for (i = 0; i < size; i++) {
+        msg = messages[i];
+        MAKE_STD_ZEVAL(zmsg);
+        object_init_ex(P_ZEVAL(zmsg), ce_kafka_message);
```
arnaud-lb: I think that this is not needed

topic.c
```diff
@@ -302,6 +302,63 @@ PHP_METHOD(RdKafka__ConsumerTopic, consume)
...
+        zend_throw_exception_ex(spl_ce_InvalidArgumentException, 0 TSRMLS_CC, "Out of range value '%ld' for $partition", partition TSRMLS_CC);
+        return;
+    }
+
+    rkmessages = malloc(sizeof(*rkmessages) * batch_size);
+
+    intern = get_kafka_topic_object(getThis() TSRMLS_CC);
+    if (!intern) {
+        return;
+    }
```
arnaud-lb, line 339: We might want to move this earlier (at least before the malloc())

topic.c
```diff
@@ -302,6 +302,63 @@ PHP_METHOD(RdKafka__ConsumerTopic, consume)
...
+
+    result = rd_kafka_consume_batch(intern->rkt, partition, timeout_ms, rkmessages, batch_size);
+
+    if (result == -1) {
+        zend_throw_exception(ce_kafka_exception, rd_kafka_err2str(rd_kafka_last_error()), err TSRMLS_CC);
```
arnaud-lb: I believe that `err` is not initialized here

message.c
nick-zh: you are totally right :smile:

topic.c
```diff
@@ -302,6 +302,63 @@ PHP_METHOD(RdKafka__ConsumerTopic, consume)
...
+        return;
+    }
+
+    if(result > 0) {
```
arnaud-lb: Should be `>= 0`, since `0` is not an error (I think ?). We want `return_value` to be an array if no error happened.

topic.c
```diff
@@ -302,6 +302,63 @@ PHP_METHOD(RdKafka__ConsumerTopic, consume)
...
+        kafka_message_list_to_array(return_value, rkmessages, batch_size TSRMLS_CC);
```
arnaud-lb: I think that we need to use `result` instead here (instead of `batch_size`), which is the number of messages consumed

topic.c
nick-zh: totally true, my bad :sweat_smile: thx for the feedback

nick-zh:
@arnaud-lb ah sry, i forgot to check my old prs for the style :sweat_smile:

topic.c
nick-zh: yes you are right :D

topic.c
```diff
@@ -302,6 +302,63 @@ PHP_METHOD(RdKafka__ConsumerTopic, consume)
...
+    }
+
+    rkmessages = malloc(sizeof(*rkmessages) * batch_size);
+
+    result = rd_kafka_consume_batch(intern->rkt, partition, timeout_ms, rkmessages, batch_size);
+
+    if (result == -1) {
+        err = rd_kafka_last_error();
+        zend_throw_exception(ce_kafka_exception, rd_kafka_err2str(err), err TSRMLS_CC);
+        return;
```
arnaud-lb, line 349: rkmessages must be freed here :)

nick-zh: added, thx, oversight from my side :+1:

nick-zh:
Since push is green (and i only added the free of messages) we can probably merge this?
PR is going to take another ~30m :D

arnaud-lb:
Thanks !

kravchenko-a:
@nick-zh you could add usage examples of the new method to the documentation.

nick-zh:
@modder2 there is an open pr for that already https://github.com/arnaud-lb/php-rdkafka-doc/pull/25

kravchenko-a:
@nick-zh there is still documentation in **this** repo in **README.md** file. I think it would be worth adding there too.

Steveb-p:
@modder2 I disagree. Keeping two sources of information can lead to conflicts and increases maintenance. Readme file is just for a quick rundown of what this extension can do.

Maybe a changelog would be a more viable option instead, if there is a need to keep track of the changes introduced.

nick-zh:
i agree with @Steveb-p, really crucial stuff (like proper shutdown, etc.) and a quick start is good to have in the readme, but otherwise it will be a pain to maintain imho and we run the danger of having outdated stuff in either one or the other place (which i want to avoid).

**Edit:** most usecases are pretty straight forward, but if you feel there are missing examples in the doc (not readme) i can add them. I have planned a to check readme and doc in the near future, to see if everything was updated accordingly and to see if stuff is missing anyway

evander-wang:
> @modder2 there is an open pr for that already [arnaud-lb/php-rdkafka-doc#25](https://github.com/arnaud-lb/php-rdkafka-doc/pull/25)

This batch examples is not exists, where can i find

nick-zh:
The doc should be enough on how to use it:
https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka-consumertopic.consumebatch.html
The rest of the code you need is same as in the low level consumer, just replace `conusme` with `consumeBatch` and the appropriate arguments:
https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.examples-low-level-consumer-basic.html
But i would advise to avoid this and use the high level consumer though without using batches.

evander-wang:
> The doc should be enough on how to use it:
> https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka-consumertopic.consumebatch.html
> The rest of the code you need is same as in the low level consumer, just replace `conusme` with `consumeBatch` and the appropriate arguments:
> https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.examples-low-level-consumer-basic.html
> But i would advise to avoid this and use the high level consumer though without using batches.

Thank you for your reply, But why the high level consumer not support batches, i want to using auto reblance and batchs, is there a way to do it ?

nick-zh:
No, there is no way of doing this for the high level consumer. In the end Kafka is for streaming data, batch consume is just one way of consuming a tad faster (fastest is through callbacks). Batch consume otherwise does not really give an advantage, if you need to process "batches" you can still abstract this in your high level logic.
Why i advise using high level consumer, it is more flexible and is not hardwired to partitions. In low level you need to change the code, etc. if you have more partitions.

evander-wang:
> No, there is no way of doing this for the high level consumer. In the end Kafka is for streaming data, batch consume is just one way of consuming a tad faster (fastest is through callbacks). Batch consume otherwise does not really give an advantage, if you need to process "batches" you can still abstract this in your high level logic.
> Why i advise using high level consumer, it is more flexible and is not hardwired to partitions. In low level you need to change the code, etc. if you have more partitions.

Thanks, low level is really inconvenient, i'm going to try high level callback now.
In addition I'm very excited to see that you are also paying attention to the swoole project. 
The coroutine is perfect, but Kafka client doesn't support it, it's a pity

Steveb-p:
@wjcgithub Nick will never advertise himself, but please take a look into https://github.com/jobcloud/php-kafka-lib if it helps you develop your application faster.

It hides some internal complexities (hopefully) that often deter newcomers.

evander-wang:
> @wjcgithub Nick will never advertise himself, but please take a look into https://github.com/jobcloud/php-kafka-lib if it helps you develop your application faster.
> 
> It hides some internal complexities (hopefully) that often deter newcomers.

Good lib, star it.

PR #259, opened 2019-08-06
Fix missing file

nick-zh:
when i generate the doc, it seems fine, but somehow the rename was not reflected on this branch @arnaud-lb


nick-zh:
closing, was fixed with last doc update

PR #263, opened 2019-08-09
Fix tp constructor args

nick-zh:


topic_partition.c
```diff
@@ -192,7 +192,10 @@ rd_kafka_topic_partition_list_t * array_arg_to_kafka_topic_partition_list(int ar
 /* {{{ proto void RdKafka\TopicPartition::__construct(string $topic, int $partition[, int $offset])
    Constructor */
 
-ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_topic_partition___construct, 0, 0, 0)
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_topic_partition___construct, 0, 0, 3)
```
arnaud-lb: Should it be 2 ?

nick-zh: you are totally right, sry i thought it's number of total args, not required :sweat_smile:

arnaud-lb:
Thanks

Issue #196, opened 2018-12-21
Any update on indefinite hangs in destructor?


jrots:
Hi, 

is there an update on the indefinite hangs in destructor when no brokers are available ? 

I tried several options but can't seem to get over this issue.. 
it's affecting my production environment atm when kafka is restarting or all the brokers are down 

 you can retry it very easily with following code: 

```
$conf = new  RdKafka\Conf();
$conf->set('socket.blocking.max.ms', '1'); // Will help librdkafka release your application to continue
$conf->set('queue.buffering.max.ms', '1'); // Dispatch asap:
$conf->set('queue.buffering.max.messages', '10'); // Wait for max 10 messages before sending to Kafka.
$conf->set("socket.timeout.ms", '10');
$conf->set('debug', 'all');
$conf->set('message.timeout.ms', '1000'); 

pcntl_sigprocmask(SIG_BLOCK, array(SIGIO));
$conf->set('internal.termination.signal', (string) SIGIO);
$producer = new RdKafka\Producer($conf);
$producer->setLogLevel(LOG_DEBUG);
$producer->addBrokers("127.0.0.1:51004,127.0.0.1:51004,127.0.0.1:51004");
$topicConfig = new RdKafka\TopicConf();
$topic = $producer->newTopic('TEST', $topicConfig);
//var_dump($topic);
try {
	$topic->produce(RD_KAFKA_PARTITION_UA, 0, 'TESTMESSAGE');
	$producer->poll(0);
} catch (\Exception $e) {
	var_dump($e);
}
die('DESTRUCT CALL HERE');
```
Afterwards rdkafka will be in a destructor loop : 

```
%7|1545411213.839|INIT|rdkafka#producer-1| [thrd:app]: librdkafka v0.11.5 (0xb05ff) rdkafka#producer-1 initialized (debug 0xffff)
%7|1545411213.839|BRKMAIN|rdkafka#producer-1| [thrd::0/internal]: :0/internal: Enter main broker thread
%7|1545411213.839|STATE|rdkafka#producer-1| [thrd::0/internal]: :0/internal: Broker changed state INIT -> UP
%7|1545411213.839|BROADCAST|rdkafka#producer-1| [thrd::0/internal]: Broadcasting state change
%7|1545411213.839|WAKEUPFD|rdkafka#producer-1| [thrd:app]: 127.0.0.1:51004/bootstrap: Enabled low-latency ops queue wake-ups
[...]

PR #264, opened 2019-08-10
expose flush

nick-zh:
@arnaud-lb i removed the poll in the destructor, because of issue https://github.com/arnaud-lb/php-rdkafka/issues/196, where the problem is, that when the broker is not available, the poll hangs forever in the destructor (because unlike for flush, there is no timeout). With flush and purge exposed, this gives the user enough flexibility to handle this on it's own i think.
Let me know what you think

Relates to: https://github.com/arnaud-lb/php-rdkafka/pull/144

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ https://arnaud-lb.github.io/php-rdkafka/phpdoc/rdkafka.examples.html
 
 ### Producing
 
+#### Creating a producer
 For producing, we first need to create a producer, and to add brokers (Kafka
 servers) to it:
 
@@ -53,6 +54,7 @@ $rk = new RdKafka\Producer($conf);
 $rk->addBrokers("10.0.0.1:9092,10.0.0.2:9092");
 ```
 
+#### Producing messages
 Next, we create a topic instance from the producer:
 
 ``` php
@@ -76,6 +78,20 @@ The first argument is the partition. RD_KAFKA_PARTITION_UA stands for
 The second argument are message flags and should always be 0, currently.
 
 The message payload can be anything.
+
+#### Proper shutdown
+This should typically be done prior to destroying a producer instance  
+to make sure all queued and in-flight produce requests are completed  
+before terminating. Use a reasonable value for $timeout_ms.
+```php
+$rk->flush($timeout_ms);
+```
+Optionally, you can purge all messages from the queue that haven't  
+been sent to the broker yet, before calling flush.
+```php
+$rk->purge(RD_KAFKA_PURGE_F_QUEUE);
+$rk->flush($timeout_ms);
+```
 
 ### High-level consuming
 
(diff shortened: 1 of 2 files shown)


arnaud-lb:
I we remove `poll()` in destruct, we don't guarantee that messages are sent anymore. what do you think about https://github.com/arnaud-lb/php-rdkafka/issues/196#issuecomment-523858096 ?

nick-zh:
@arnaud-lb i think we can remove poll in destruct, i updated the readme and doc examples accordingly

arnaud-lb:
Thanks !

anam-hossain:
@nick-zh An example would be great. Thanks for the PR. We needed this.

nick-zh:
@anam-hossain no worries, there is a very small example here: https://arnaud.le-blanc.net/php-rdkafka/phpdoc/rdkafka.examples-producer.html and also in the readme: https://github.com/arnaud-lb/php-rdkafka#proper-shutdown
Let me know if something is unclear or if i can improve the docs

anam-hossain:
@nick-zh Thanks. The docs look good. However, for the beginner, it is difficult to understand when to use purge and flush. Furthermore, how to handle the failure. A complete example will help us a lot.

nick-zh:
@anam-hossain thank you for the feedback, i added some more info to the doc, until that is merged, i think a good example would probably look a little bit like this:
```
$conf = new RdKafka\Conf();
$conf->set('metadata.broker.list', 'localhost:9092');

//If you need to produce exactly once and want to keep the original produce order, uncomment the line below
//$conf->set('enable.idempotence', 'true');

$topic = $producer->newTopic("test");

for ($i = 0; $i < 10; $i++) {
    $topic->produce(RD_KAFKA_PARTITION_UA, 0, "Message $i");
    $producer->poll(0);
}

for ($flushRetries = 0; $flushRetries < 10; $flushRetries++) {
    $result = $producer->flush(10000);
    if (RD_KAFKA_RESP_ERR_NO_ERROR === $result) {
        break;
    }
}

if (RD_KAFKA_RESP_ERR_NO_ERROR !== $result) {
    throw new \RuntimeException('Was unable to flush, messages might be lost!');
}
```
I think most people won't make use of `purge`. When you use `produce`, messages land in an internal queue before being sent to Kafka. With `purge` you can empty that internal queue (or even kill inflight messages). I think most people want their messages sent once they used `produce`.
Let me know if you need more clarification on something.

Steveb-p:
@nick-zh I'm pretty certain `flush` does not need to be in a loop. At least I'm wondering if it has to :thinking: 

https://docs.confluent.io/3.1.1/clients/librdkafka/rdkafka_8h.html#aaff06c4372bce917c17f3c1a5d8b205d

nick-zh:
@Steveb-p the problem is, flush can return `RD_KAFKA_RESP_ERR__TIMED_OUT` if it wasn't successful, to be on the safe side, you would want to retry that (it mostly won't happen i hope).
Of course you could also just do:
```
$result = $producer->flush(10000);

if (RD_KAFKA_RESP_ERR_NO_ERROR !== $result) {
    throw new \RuntimeException('Was unable to flush, messages might be lost!');
}
```

anam-hossain:
@nick-zh Thank you very much. Much appreciated.

Issue #248, opened 2019-07-27
Cleanup deprecated librdkafka functions


nick-zh:
This is a ticket to track progress on deprecating or replacing deprecated librdkafka functions and providing their alternatives:

- [x] deprecate ` rd_kafka_errno` (#254)
- [x] deprecate `rd_kafka_errno2err` (#254)
- [x] replace internal `rd_kafka_errno2err` calls with  `rd_kafka_last_error()` (#252) 
- [x] deprecate `rd_kafka_set_log_level` (#251) 
- [x] deprecate `rd_kafka_set_logger` (#251) 
- [x] deprecate `rd_kafka_conf_set_default_topic_conf` (#266)
- [x] expose  `rd_kafka_conf_set_log_cb` (#253) (related issues #81, #127,#233)

arnaud-lb:
+1.

Regarding `rd_kafka_last_error()`, we should probably not expose it, since we check it ourselves already (we throw exceptions in case of errors). We should replace `rd_kafka_errno2err(errno)` calls by `rd_kafka_last_error()` calls internally, though.

Regarding `rd_kafka_set_log_level()`, we should deprecated `Rdkafka::setLogLevel()` entirely (this method is the OO-equivalent of `rd_kafka_set_log_level()`).

Same for `rd_kafka_set_logger()`.

nick-zh:
@arnaud-lb thanks for the feedback, i will try to implement the changes accordingly :+1:

nick-zh:
@arnaud-lb is there a way to mark `rd_kafka_errno2err` and `rd_kafka_errno` as deprecated? I was unable to find if this is possible with `PHP_FE` since i was unable to find it in the internals doc :sweat_smile:

arnaud-lb:
Yes, you can use `PHP_DEP_FE()`; it's an alias for `ZEND_DEP_FE()`: https://github.com/php/php-src/blob/php-7.3.7/Zend/zend_API.h#L74

PR #266, opened 2019-08-10
deprecate set default topic conf

nick-zh:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2095
#248


arnaud-lb:
Thanks

Issue #241, opened 2019-07-22
Add kafka version to issue template


nick-zh:
Dear @arnaud-lb 

I think we should add the `kafka version:` to the issue template as well to spot faster
if a feature is trying to be used, that might not be available.

Cheers,
Nick

PR #268, opened 2019-08-11
update issue template

nick-zh:
@arnaud-lb i took the liberty of dropping the google groups link, since it seems unmaintained. I think the less channels, the better, also i really like github, since you can search if other people had the same issue in a very easy way. I added the kafka version and adjusted the template a little bit, hope it is ok, otherwise, let me know.
https://github.com/arnaud-lb/php-rdkafka/issues/241


arnaud-lb:
+1

Thanks !

PR #270, opened 2019-08-24
add offsetsfortimes for high level consumer

nick-zh:
@arnaud-lb i think this makes sense, a consumer might want to resume from a certain time frame, which would be possible using `offsetsForTimes` and `assign`


arnaud-lb:
Thanks !

PR #271, opened 2019-08-24
add querWatermarkOffsets for high level consumer

nick-zh:
@arnaud-lb i think this makes sense to add, so a high level consumer is able to track his own lag with `queryWatermarkOffsets` and `offsetPositions`


arnaud-lb:
Thanks !

Issue #272, opened 2019-08-25
Segfault on Producer::newQueue


nick-zh:
* PHP version: 7.3
 * librdkafka version: 1.1.0
 * php-rdkafka version: master
 * kafka version: -

It seem when we call `newQueue` it results in a segfault right now. gdb says it happens on this line:
`zend_hash_index_add_ptr(&intern->queues, (zend_ulong)queue_intern, queue_intern);`
**Update:** Root cause is, that the queues only get initialized for consumers atm, fix is on the way

PR #273, opened 2019-08-25
fix segfault on Producer::newQueue

nick-zh:
https://github.com/arnaud-lb/php-rdkafka/issues/272

### DIFF
--- a/rdkafka.c
+++ b/rdkafka.c
@@ -281,6 +281,58 @@ PHP_METHOD(RdKafka__Consumer, __construct)
 }
 /* }}} */
 
+/* {{{ proto RdKafka\Queue RdKafka\Consumer::newQueue()
+   Returns a RdKafka\Queue object */
+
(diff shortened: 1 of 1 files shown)


nick-zh:
@arnaud-lb let me know if this is ok, first i wanted to initialize queue always, but since it seems this feature is only thought for consumers, i decided to return a `RD_KAFKA_RESP_ERR__NOT_IMPLEMENTED` instead, since `newQueue` is not meant for producers. Let me know what you think.

arnaud-lb:
Could we remove the method from `RdKafka\Kafka`, and only add it on `RdKafka\Consumer` ? Otherwise, we should throw an exception (returning a non-`RdKafka\Queue` might be confusing / unexpected).

nick-zh:
@arnaud-lb totally, i missed the consumer and producer blocks, but this of course makes way more sense actually :+1:

arnaud-lb:
Thanks !

Issue #275, opened 2019-09-05
php core dump from ext-rdkafka


daydaygo:
* PHP version: 7.1.13
 * librdkafka version: 0.11.4
 * php-rdkafka version: 3.1.2
 * kafka version: 2.11-0.10.1.0

use `gdb php /path/to/core.php`, then use `bt` for backtrace

```
#0  ZEND_BIND_STATIC_SPEC_CV_CONST_HANDLER () at /opt/phpbrew/build/php-7.1.13/Zend/zend_vm_execute.h:40686
#1  0x000000000075139b in execute_ex (ex=<optimized out>) at /opt/phpbrew/build/php-7.1.13/Zend/zend_vm_execute.h:429
#2  0x00000000006fce51 in zend_call_function (fci=0x7f8d8d5974e0, fci_cache=0x7f8d8d597518)
    at /opt/phpbrew/build/php-7.1.13/Zend/zend_execute_API.c:855
#3  0x00007f8d913cf7b9 in zend_string_alloc (persistent=0, len=140734397967328)
    at /opt/phpbrew/php/php-7.1.13/include/php/Zend/zend_string.h:124
#4  zend_string_init (persistent=0, len=140734397967328, str=0x2ebcdb0 "")
    at /opt/phpbrew/php/php-7.1.13/include/php/Zend/zend_string.h:158
#5  kafka_conf_error_cb (rk=<optimized out>, err=<optimized out>, reason=0x2ebcdb0 "", opaque=0x7fff47cc6bc0)
    at /opt/phpbrew/build/php-7.1.13/ext/rdkafka/conf.c:154
#6  0x0000000002ebcdd8 in ?? ()
#7  0xffffffffffffffff in ?? ()
#8  0x0000000000000000 in ?? ()
```

nick-zh:
@daydaygo can you provide more information on what actually led to this core dump? thx in advance

daydaygo:
- gdb
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/opt/phpbrew/php/php-7.1.13/bin/php /opt/pt-php-project/script/kafka-consumer/p'.
Program terminated with signal 11, Segmentation fault.
#0  ZEND_BIND_STATIC_SPEC_CV_CONST_HANDLER () at /opt/phpbrew/build/php-7.1.13/Zend/zend_vm_execute.h:40686
[...]

Issue #157, opened 2018-03-15
Cannot trigger callbacks in high level consumer


beriba:
* PHP version: 5.6
 * librdkafka version: 0.11.0-RC1
 * php-rdkfaka version: 3.0.5

In documentation for class RdKafka\KafkaConsumer I cannot see any possibility to trigger manually callbacks, especially setStatsCb(). Such option is available in low level consumer by using poll(). In high level consumer the only way I see to trigger callbacks is to execute consume() but it's not an option due to bussiness requirements. Can you provide an equivalent of poll() method to the RdKafka\KafkaConsumer class?

nick-zh:
@beriba may i ask what our usecase is? Since you don't want to consume you could just do this with the low level consumer.

beriba:
Since it was 1.5 years ago I don't remember what was the use case. Feel free to close this issue because I certainly no longer need it.

PR #276, opened 2019-09-05
remove poll

nick-zh:
#275

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -77,6 +77,12 @@ if test "$PHP_RDKAFKA" != "no"; then
     AC_MSG_WARN([murmur2 partitioner is not available])
   ])
 
+  AC_CHECK_LIB($LIBNAME,[rd_kafka_destroy_flags],[
+    AC_DEFINE(HAS_RD_KAFKA_DESTROY_FLAGS,1,[ ])
+  ],[
+    AC_MSG_WARN([Destroy flags are not available])
+  ])
+
   LDFLAGS="$ORIG_LDFLAGS"
   CPPFLAGS="$ORIG_CPPFLAGS"
 
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -53,13 +53,11 @@ static void kafka_consumer_free(zend_object *object TSRMLS_DC) /* {{{ */
 
     if (intern->rk) {
         err = rd_kafka_consumer_close(intern->rk);
+        
         if (err) {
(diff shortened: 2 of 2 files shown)


nick-zh:
Salut @arnaud-lb 

I think this is related to #196 since the core dump happens during the destruct as well (i might be wrong though). I removed poll for high level as well, in my PR i only did it for low level producer / consumer.

From what i read here:
https://github.com/edenhill/librdkafka/wiki/Proper-termination-sequence#producers-and-simple-legacy-consumers
It seems poll is only used for those two for proper shutdown anyway.
In the high level section:
https://github.com/edenhill/librdkafka/wiki/Proper-termination-sequence#high-level-kafkaconsumer
It is only mentioned, that we need to call `rd_kafka_consumer_close` and `rd_kafka_destroy` which we do.

I found that `rd_kafka_consumer_close` already does a poll itself:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.c#L2873
From the comment here, i see that rebalance_cb or others could still be called if set:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2480

Maybe this is unsafe to do in the destructor, what do you think? It seems this exactly what is happening in the issue mentioned, callback is called but killed during execution time.

I think from what i gathered we could avoid the callbacks being called by setting the destroy flag properly:
```
rd_kafka_destroy_flags(intern->rk, RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE);
```
`rd_kafka_destroy_flags` also calls `rd_kafka_consumer_close` but sets the flag accordingly beforehand.

Reference: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2516

nick-zh:
I did some more digging and found that `rd_kafka_consumer_close` already does a poll itself:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.c#L2873
Which is actually blocking, so i think we should be fine.

nick-zh:
@arnaud-lb maybe it would be worth to expose poll for the high level consumer (`rd_kafka_poll_cb`), it was also requested in #157 a while back. I think if we are really fast and the last callback is triggered and we already destroy stuff, it could still happen, that the last callback sort of dies (not sure).
What do you think?

nick-zh:
closing this since doing it as in #278 doesn't rely on a specific version of librdkafka, since any version prior to `0.11.6` is not able to do it in the way proposed here

PR #281, opened 2019-09-16
set log.queue in conf itself

### LINKED ISSUE #280: Segmentation fault with Conf::setLogCb
* PHP version: 7.2.21 & 7.4-rc1 (both with xdebug)
 * librdkafka version: v1.0.0
 * php-rdkafka version: master
 * kafka version: -

Script crashes with "segmentation fault" while all works fine when explicitly setting ```$conf->set('log.queue', 'true');```:

```php
<?php

use RdKafka\Conf;
use RdKafka\Consumer;

$loggerCallbacks = 0;

$conf = new Conf();
$conf->set('debug', 'consumer');
$conf->set('log_level', (string)LOG_DEBUG);
//$conf->set('log.queue', 'true');
$conf->setLogCb(function (Consumer $consumer, int $level, string $fac, string $buf) use (&$loggerCallbacks) {
    $loggerCallbacks++;
});

$consumer = new Consumer($conf);
do {
    $consumer->poll(0);
} while ($loggerCallbacks === 0);

var_dump($loggerCallbacks); // expect 1
```

nick-zh:
fixes #280
was only done for low level consumer, so i moved the set to conf directly, which makes more sense anyway

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -720,6 +720,7 @@ PHP_METHOD(RdKafka__Conf, setLogCb)
     zend_fcall_info fci;
     zend_fcall_info_cache fcc;
     kafka_conf_object *conf;
+    char errstr[512];
 
     if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "f", &fci, &fcc) == FAILURE) {
         return;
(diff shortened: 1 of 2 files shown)


conf.c
```diff
@@ -742,6 +743,7 @@ PHP_METHOD(RdKafka__Conf, setLogCb)
     conf->cbs.log->fcc = fcc;
 
     rd_kafka_conf_set_log_cb(conf->u.conf, kafka_conf_log_cb);
+    rd_kafka_conf_set(conf->u.conf, "log.queue", "true", errstr, sizeof(errstr));
```
dirx, line 746: in case log.queue will be set to "false" after setLogCb - remove callback again?

nick-zh: i am not so sure, this feels a bit like magic, i mean if somebody does this, you are like sabotaging yourself :smile:
i totally get your point, but i am not sure if it is worth to cover this, @arnaud-lb thoughts?

arnaud-lb:
Thanks

Issue #289, opened 2019-10-09
rdkafka 4.0 questions


Jukit-Barrett:
* PHP version: docker  php:7.2.23-cli-alpine
 * librdkafka version: 
 * php-rdkafka version: 4.0.0
 * kafka version: 

## question 1:
````
/usr/src/php/ext/rdkafka/rdkafka.c: In function 'zim_RdKafka__Kafka_setLogger':
/usr/src/php/ext/rdkafka/rdkafka.c:728:5: warning: 'rd_kafka_set_logger' is deprecated [-Wdeprecated-declarations]
     rd_kafka_set_logger(intern->rk, logger);
     ^~~~~~~~~~~~~~~~~~~
In file included from /usr/src/php/ext/rdkafka/php_rdkafka.h:24,
                 from /usr/src/php/ext/rdkafka/rdkafka.c:28:
/usr/include/librdkafka/rdkafka.h:3792:6: note: declared here
 void rd_kafka_set_logger(rd_kafka_t *rk,
      ^~~~~~~~~~~~~~~~~~~
````

## question 2:
````
/usr/src/php/ext/rdkafka/conf.c: In function 'zim_RdKafka__Conf_setLogCb':
/usr/src/php/ext/rdkafka/conf.c:759:44: warning: passing argument 2 of 'rd_kafka_conf_set_log_cb' from incompatible pointer type [-Wincompatible-pointer-types]
     rd_kafka_conf_set_log_cb(conf->u.conf, kafka_conf_log_cb);
                                            ^~~~~~~~~~~~~~~~~
In file included from /usr/src/php/ext/rdkafka/php_rdkafka.h:24,
                 from /usr/src/php/ext/rdkafka/conf.c:26:
/usr/include/librdkafka/rdkafka.h:1733:13: note: expected 'void (*)(const rd_kafka_t *, int,  const char *, const char *)' {aka 'void (*)(const struct rd_kafka_s *, int,  const char *, const char *)'} but argument is of type 'void (*)(rd_kafka_t *, int,  const char *, const char *)' {aka 'void (*)(struct rd_kafka_s *, int,  const char *, const char *)'}
      void (*log_cb) (const rd_kafka_t *rk, int level,
      ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                           const char *fac, const char *buf));
                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #290, opened 2019-10-09
fix argument

nick-zh:
#289


nick-zh:
Hey @arnaud-lb, should not be a huge problem, but it gets rid of a warning during make, i missed the `const` in the callback declaration (https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_conf.c#L2429), so i think this fix should be fine. Let me know

arnaud-lb:
Thanks !

PR #291, opened 2019-10-09
add test suite in pecl archive

remicollet:
This also rename package2.xml (deprecated) to package.xml

`pecl package` will use it.


nick-zh:
@remicollet thank you very much
CC: @arnaud-lb

remicollet:
FYI, this is to make easier for "downstream" to run test suite during package build.

arnaud-lb:
Thanks @remicollet !

PR #303, opened 2019-10-26
add slack channel

nick-zh:
@arnaud-lb while i mostly prefer communication over issues (searchable, history, etc.) sometimes it would really be nice to have a quick chat to sort stuff out.
That's why i created a php-kafka workspace, i think colaboration is nice and important also with other people using the ext (and in the future ffi stuff).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # PHP Kafka client - php-rdkafka
 
+[![Join the chat at https://gitter.im/arnaud-lb/php-rdkafka](https://badges.gitter.im/arnaud-lb/php-rdkafka.svg)](https://gitter.im/arnaud-lb/php-rdkafka?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)  
+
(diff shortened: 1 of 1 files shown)


arnaud-lb:
I'm using slack on other projects, and while it's convenient, the lack of history is painful. It's not rare that I search for a conversation from a few days ago, only to find that it's gone.

I would really prefer an alternative with unlimited history.

nick-zh:
@arnaud-lb i totally agree regarding the history
Since i am a heavy slack user though, i think this would be a nice to have for discussions, related to kafka but maybe unrelated to rdkafka, that's why i created a slack channel with a more general purpose.
For me mostly, issues should be here and not in a chat anyway (search, google search, etc.).
I think for php-rdakfa itself, it is totally fine to use gitter.
I don't want to push anything, if you are not happy with adding slack in the readme as an alternative, i am totally fine with it and you can close the PR of course :+1:

Steveb-p:
Symfony is using Slack, but they do have a lot bigger community and a company backing it up I guess.

We could also use Discord, since while it is designed for gamers mostly it does support most of Slack functionality and is used for example by React community.

PR #308, opened 2019-11-04
Add .editorconfig

Steveb-p:
See https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+insert_final_newline = true
+
+[.travis.yml]
+indent_style = space
+indent_size = 4
+
+[*.md]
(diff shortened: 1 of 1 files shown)


arnaud-lb:
Thanks ! Should we add rules for *.c / *.h as well ?

Steveb-p:
> Thanks ! Should we add rules for *.c / *.h as well ?

I wasn't sure what you're preferences in terms of C files are, so I've left it out. I'm also not aware of any standard that is more common or accepted, unlike PSR's in PHP.

If you'd like it in any specific way, just tell me and I'll update the PR with those settings.

arnaud-lb:
Thanks ! I've merged and added some rules for *.{c,h}

Issue #242, opened 2019-07-24
Expose rd_kafka_consume_callback()


nick-zh:
According to librdkafka:
```
`rd_kafka_consume_callback()` provides higher throughput performance
 than both `rd_kafka_consume()` and `rd_kafka_consume_batch()`
```

PR #310, opened 2019-11-08
Add kafka consume callback

nick-zh:
#242

### DIFF
--- a/rdkafka.c
+++ b/rdkafka.c
@@ -39,6 +39,8 @@
 #include "topic_partition.h"
 #include "fun.h"
 
+extern void (*consume_callback)(void);
+
 #if RD_KAFKA_VERSION < 0x000b0000
 #	error librdkafka version 0.11.0 or greater required
 #endif
@@ -249,6 +251,7 @@ PHP_METHOD(RdKafka, __construct)
 }
 /* }}} */
 
+
 /* {{{ proto RdKafka\Consumer::__construct([RdKafka\Conf $conf]) */
 
 ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_consumer___construct, 0, 0, 0)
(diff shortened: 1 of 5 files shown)


rdkafka.c
```diff
@@ -273,6 +275,50 @@ PHP_METHOD(RdKafka__Consumer, __construct)
...
 
+/* {{{ proto RdKafka\Consumer::consumeCallback([RdKafka\Topic $topic, int $partition, int timeout_ms]) */
+
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_consumer__consume_callback, 0, 0, 3)
+    ZEND_ARG_INFO(0, topic)
+    ZEND_ARG_INFO(0, partition)
+    ZEND_ARG_INFO(0, timeout_ms)
+ZEND_END_ARG_INFO()
+
+PHP_METHOD(RdKafka__Consumer, consumeCallback)
```
arnaud-lb: We should try to keep the API close to rdkafka's: The callback parameter should be a parameter of this method, not a config parameter :)

Also, since the original function acts on a topic (the first param is a topic), the method should be on `ConsumerTopic`.

rdkafka.c
```diff
@@ -273,6 +275,50 @@ PHP_METHOD(RdKafka__Consumer, __construct)
...
+        zend_throw_exception(ce_kafka_exception, "You need to set a consume callback", 0 TSRMLS_CC);
+        return;
+    }
+
+    topic = get_kafka_topic_object(ztopic TSRMLS_CC);
+    if (!topic) {
+        return;
+    }
+
+   RETURN_LONG(rd_kafka_consume_callback(topic->rkt, partition, timeout_ms, consume_callback, rd_kafka_opaque(intern->rk)));
```
arnaud-lb: Here, we could pass the PHP callback as opaque, and `consume_callback` should be a static function that just casts the opaque as a PHP callback, and call it :) (no need to use `interns->cbs`)

When rd_kafka_consume_callback returns, we can free the PHP callback.

rdkafka.c
```diff
@@ -273,6 +275,50 @@ PHP_METHOD(RdKafka__Consumer, __construct)
...
+        return;
+    }
+
+    intern = get_kafka_object(getThis() TSRMLS_CC);
+    if (!intern) {
+        return;
+    }
+
+    if (!intern->cbs.consume) {
+        zend_throw_exception(ce_kafka_exception, "You need to set a consume callback", 0 TSRMLS_CC);
```
arnaud-lb: We should avoid "you" in messages :) See "Personalization" on http://doc.php.net/tutorial/style.php

nick-zh:
@arnaud-lb thx for the feedback, i tried to adjust accordingly. I am a bit unhappy though, that we pass the callback store and free it after every call. I am not sure how much overhead this generates, but since this method is supposed to be the fastest of all, i am wondering, if it would not be nicer to set it once and then being able to always use the same, wdyt?
Also this PR is not ready yet, because as of now, i am unsure, how it is intended to commit in this usecase, there is an open question for this https://github.com/edenhill/librdkafka/issues/2617

topic.h
```diff
@@ -29,6 +29,11 @@ typedef struct _kafka_topic_object {
 #endif
 } kafka_topic_object;
 
+typedef struct _php_callback {
```
nick-zh: was unsure what a proper naming should be, i wanted to reuse the kafka_conf_callback, but the naming is not fitting, maybe we should rename kafka_conf_callback to something more general?

arnaud-lb: This struct looks good to me. Maybe just declare it in `topic.c`, since it's going to be used only there.

topic.c
```diff
@@ -83,6 +107,52 @@ kafka_topic_object * get_kafka_topic_object(zval *zrkt TSRMLS_DC)
...
+
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_consume_callback, 0, 0, 3)
+    ZEND_ARG_INFO(0, partition)
+    ZEND_ARG_INFO(0, timeout_ms)
+    ZEND_ARG_INFO(0, callback)
+ZEND_END_ARG_INFO()
+
+PHP_METHOD(RdKafka__ConsumerTopic, consumeCallback)
+{
+    php_callback *cb;
```
arnaud-lb: This can be allocated on the stack, so that you don't need to allocate / free it. 
This also allows to pass cb.fci and cb.fcc directly to `zend_parse_parameters()`.

```
php_callback cb;

if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "llf", &partition, &timeout_ms, &cb.fci, &cb.fcc) == FAILURE) {
```

nick-zh: oh cool, thx for the feedback, i will check it out :+1:

arnaud-lb:
Thanks !

PR #315, opened 2019-11-29
fix abstract class topic

### LINKED ISSUE #314: RdKafka\Topic is not found in PHP7.4
Discovered during https://github.com/jobcloud/messaging-lib/pull/65
phpstan complained with:
```
Class RdKafka\Topic was not found while trying to analyse it - autoloading is probably not configured properly.
```

nick-zh:
should fix #314


arnaud-lb:
Thanks !

PR #317, opened 2019-11-29
Update readmemd installation guide

BOR4:
I see lot of people (including myself) accidentally end up with old librdkafka version. Extended guide with more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,23 @@ The source of the documentation can be found [here](https://github.com/arnaud-lb
 
 ## Installation
 
+### Installing [librdkafka](https://github.com/edenhill/librdkafka) dependency (Ubuntu and Debian)
+
+:warning: To install the latest stable version make sure to first follow these [instructions].(https://docs.confluent.io/current/installation/installing_cp/deb-ubuntu.html#get-the-software)
+
+If you install by running:
+```bash
+apt-get -y install librdkafka-dev
+```
+librdkafka version 0.x.x will be installed.
+
+You can check currently installed version of librdkafka-dev by running:
+```bash
+apt-cache policy librdkafka-dev
+```
+
+### Instaling php-rdkafka
+
 https://arnaud-lb.github.io/php-rdkafka/phpdoc/rdkafka.setup.html
 
 ## Examples


Steveb-p:
:heart: Debian/Ubuntu repositories never updating those dependencies.

BOR4:
@Steveb-p we spent good 4 hours wondering what was going on while production was on fire :). Let's save others!

Steveb-p:
@BOR4 just to note docs that you linked provide the whole package of confluent. The only thing that needs to be done for it work (librdkafka 1.0) is adding the repository.

```
wget -qO - https://packages.confluent.io/deb/5.3/archive.key | sudo apt-key add -
sudo add-apt-repository "deb [arch=amd64] https://packages.confluent.io/deb/5.3 stable main"
sudo apt-get update
sudo apt-get install -y librdkafka1
```
see: https://docs.confluent.io/current/installation/available_packages.html#librdkafka

BOR4:
that is true. It can be written like that as well. Maybe nicer as it keeps everything in 1 place.

nick-zh:
@BOR4 thanks for bringing this to our attention, i would like to keep everything in one place instead of splitting it with a part in the readme and a part in the doc.
I think it would be good to adjust the [requirements part](https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.requirements.html) accordingly. Maybe with some examples for docker, debian, etc. and with a big hint, that most distros come with a very outdate lib version
I can take care of that or you can open a PR for the doc, whatever you prefer.

BOR4:
@nick-zh I am really bad at writing guides and docs. If you would be so kind and take this over yourself.

PS My 2nd pull request will be about requirement to call flush since version 4.0.0 (right now it is just a suggestion in readmemd).

nick-zh:
@BOR4 i will gladly do so, we improved this recently, e.g.: https://github.com/arnaud-lb/php-rdkafka#proper-shutdown
but if there are parts where this seems optional, please do correct and improve them, thx in advance :heart:

BOR4:
thanks for all the work you've put into this project @nick-zh . I will check it out when I catch time!

nick-zh:
thanks for taking the time and the kind words. i will improve the doc today / this weekend :+1:

Steveb-p:
> thanks for all the work you've put into this project @nick-zh . I will check it out when I catch time!

Yeah @nick-zh definitely deserves a pat on the back :smile: and a bottle of good whiskey :champagne:

PR #318, opened 2019-12-02
remove header destroy

nick-zh:
was reported by @Steveb-p during testing, i was able to reproduce that and it seems,
that it fails (core dumped) when doing:
```
rd_malloc
```
somewhere, from what i understood, it is because the produce_request is not created yet and i freed the header to early. The error happens shortly after freeing the headers (not immediately) this is what lead to my conclusion.

### DIFF
--- a/topic.c
+++ b/topic.c
@@ -619,6 +619,7 @@ PHP_METHOD(RdKafka__ProducerTopic, producev)
     );
 
     if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
+        rd_kafka_headers_destroy(headers);
         zend_throw_exception(ce_kafka_exception, rd_kafka_err2str(err), err TSRMLS_CC);
         return;
     }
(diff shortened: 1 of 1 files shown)


Steveb-p:
Tests from #223 that run integration testing now indicate that `producev` method behaves correctly with this cherry-picked :+1:

arnaud-lb:
From the docblock of [`RD_KAFKA_V_HEADERS()`](https://github.com/edenhill/librdkafka/blob/51d952e51ae879e293eeda45e9a5639b58ee54f5/src/rdkafka.h#L1029), it appears that we lose ownership of the `rd_kafka_headers_t` after calling `producev`, unless it failed.

If true, this means that we should free `headers` if `producev` returns an error (but only in this case).

nick-zh:
@arnaud-lb yeah i thought so, yesterday night i did some more digging and found this, i will adapt accordingly, thx a lot

arnaud-lb:
Thanks !

Issue #320, opened 2019-12-04
Remove partition 0 from RD_KAFKA_RESP_ERR__TIMED_OUT


nick-zh:
The message object contains partition: 0 when we have `RD_KAFKA_RESP_ERR__TIMED_OUT` in the high level consumer.
This should be fixed to null to avoid confusion about the consumer being stuck on one partition.
Reported by @alexandra-default

PR #321, opened 2019-12-04
dont set partition to zero if null

nick-zh:
#320


Steveb-p:
Travis failed to install php 7.0... :smile:

nick-zh:
yeah this happens rarely, but short job restart always helps :smile:

arnaud-lb:
Thanks !

PR #322, opened 2019-12-09
Issue warns

remicollet:
Remaining warnings: [-Wdeprecated-declarations]

* `rd_kafka_errno2err`
* `rd_kafka_set_logger`

About [-WFormat]

The call to zend_parse_parameter for l have to use a `zend_long `in 7+ (not a `long`), and for consistency use `ZEND_LONG_FMT `(not `%ld`)

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -823,7 +823,7 @@ PHP_METHOD(RdKafka__TopicConf, setPartitioner)
             break;
 #endif
         default:
-            zend_throw_exception_ex(spl_ce_InvalidArgumentException, 0 TSRMLS_CC, "Invalid partitioner given", partitioner TSRMLS_CC);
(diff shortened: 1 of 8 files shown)


nick-zh:
@remicollet thank you! @arnaud-lb please have a look

nick-zh:
@remicollet regarding the deprecation warnings, is there something i can do to help fix this?
the problem is the functions `setLogger` and `rd_kafka_errno2err` are deprecated (as well as their librdkafka calls, that's why i deprecated them), but i don't think we can remove them for now.

remicollet:
As the PHP side is also deprecated, I think everything is OK, and these warnings can be ignored

remicollet:
The 4th commit switch to zend_long for zend_parse_parameters arg with compat macro for 5.6

arnaud-lb:
Thanks !

Issue #324, opened 2019-12-12
Error on commit from Message, partition is null instead of zero


Triasto:
* PHP version: 7.3
 * librdkafka version: 1.3.0
 * php-rdkafka version:4.0.1

Yesterday we updated our images and we encountered a whole lot of errors and Messages couldn't be committed anymore.

We produce our data as stated in:
https://github.com/arnaud-lb/php-rdkafka/blob/master/examples/producer.php

Our guess is that the following changes somehow breaks our code.

https://github.com/arnaud-lb/php-rdkafka/pull/321

When we dump the message we see that the consumed message has a partition of null instead of an integer.

nick-zh:
@Triasto thx for the report, meanwhile i suggest to stay at `rdkafka:4.0.0` i will check this out tonight

nick-zh:
Short update, i can confirm that this is an acutal bug, it seems somehow for partition zero i also get null (really unexpected), at least in the few minutes i had to check, which is not cool, because this will cause the behaviour mentioned. I dig a bit more and i will provide a fix this evening. Sry for the inconvenience

Triasto:
Thanks for checking this so fast!

renatomefi:
The version 4.0.1 could be marked as unstable and possibly removed from pecl, this way people jump directly to the next one!

Also as this is a patch that's including features a good ideas would've been releasing as 4.1.0.

Thanks for maintaining it, the fact people are finding this issues shows there's a traction going on here and getting a bit more of criteria on releases will make everyone's lives better!

nick-zh:
@arnaud-lb anyway we can do this on pecl? I added a warning to the release notes, that's the best i was able to do for now.
@renatomefi yeah totally agree, i want to commit some time to add more tests, sadly i didn't have much time recently, but this shows how important it is.

Steveb-p:
@nick-zh I think we also need to reiterate on the testing process. The current time it takes to complete is just way too much considering the test matrix we have. We should prioritize some crucial test scenarios and run the others only if they pass. This could mean difference between running automatic tests for 15 minutes instead of 2 hours (which - mind you - block the whole travis pipeline for the repository :disappointed: )

arnaud-lb:
#327 fixes this.

I've released 4.0.2 on PECL: https://pecl.php.net/package/rdkafka

Thanks @nick-zh.

nick-zh:
Awesome @arnaud-lb thx! Thanks too for taking the time during this busy season.

PR #325, opened 2019-12-12
fix partition for consumption

nick-zh:
#324


nick-zh:
@arnaud-lb i will still dig into this a bit more, it seems for partition zero, i will also get null, which strikes me as quite odd, so i want to check if there is something else going wrong that i could fix, i don't like this fix too much :smile:

nick-zh:
@arnaud-lb ok i checked, either i am missing something or this is really strange, even if i change it to
```
if (NULL != message->partition) {
```
which should be basically the same as we had (since partition is int32_t), i end up with `NULL` for partition zero. Please let me know if i missed something.
Would be really nice if we roll this out kind of fast, since this is kind of a critical bug.

nick-zh:
was fixed with #327 , will be reverted back to initial behaviour

PR #326, opened 2019-12-15
Travis docker

nick-zh:


nick-zh:
i think this setup could improve build time, somehow the tests have leaks though:
```
==78== Invalid free() / delete / delete[] / realloc()
==78==    at 0x4C98A62: free (vg_replace_malloc.c:530)
==78==    by 0x40587E5: ??? (in /lib/ld-musl-x86_64.so.1)
==78==  Address 0x4ea0180 is in a rw- mapped file /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so segment
==78== 
```
Not sure if this not because of the alpine image that uses musl, but in anycase i was unable to surpress it with
```
Memcheck: Free
```
or with `--run-libc-freeres=no`.
I am going to check next, if i can really verify that there is a duplicate free.
Was also unable to generate the suppression for this leaks...

nick-zh:
found it, the alias was not used properly, was able to fix it

.travis/test.sh
```diff
@@ -1,14 +1,18 @@
...
 set -xve
 
-export PATH=$TRAVIS_BUILD_DIR/.travis:$PATH
-
-showmem=
-if grep -q 'cfgfiles.*mem' run-tests.php; then
+if grep -q 'cfgfiles.*mem' ${TRAVIS_BUILD_DIR}/run-tests.php; then
     echo "Will enable the --show-mem flag"
-    showmem=--show-mem
+    exprt SHOWMEM=--show-mem
```
Steveb-p: Typo here (`exprt`)

nick-zh: haha seems everybody is busy on a sunday :smile: thx man

.travis/valgrind
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-/usr/bin/valgrind --suppressions="$TRAVIS_BUILD_DIR/.travis/default.supp" "$@"
+/usr/bin/vvalgrind --suppressions="/php-rdkafka/.travis/default.supp" "$@"
```
Steveb-p, line 2: I'm pretty sure changing to `vvalgrind` is not intentional and there won't be a binary with this name?

nick-zh: it is, the old way (having a binary in ./travis or setting an alias) did not work in the image, so to be able to use a suppress file, i am creating a wrapper in /usr/bin and moving the original to `vvalgrind`.

nick-zh: the nicest way would actually be to be able to pass arguments to `run-tests.php` for valgrind, but sadly that is not possible (or at least i was unable to find it) :disappointed:

.travis/default.supp
```diff
@@ -3,3 +3,9 @@
     Memcheck:Cond
     fun:zend_string_equal_val
 }
+
+{
+    Surppress_duplicate_free
```
Steveb-p, line 8: Isn't this a typo?

PR #327, opened 2019-12-15
Revert "dont set partition to zero if null"

arnaud-lb:
Reverts arnaud-lb/php-rdkafka#321

### DIFF
--- a/message.c
+++ b/message.c
@@ -56,9 +56,7 @@ void kafka_message_new(zval *return_value, const rd_kafka_message_t *message TSR
     if (message->rkt) {
         zend_update_property_string(NULL, return_value, ZEND_STRL("topic_name"), rd_kafka_topic_name(message->rkt) TSRMLS_CC);
     }
-    if (message->partition) {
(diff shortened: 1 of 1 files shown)


nick-zh:
i think ultimately, we want #325 though, since always receiving partition 0 for timed out is a bit strange, wdyt?
Relates to #324

arnaud-lb:
If librdkafka returns partition 0, we should expose exactly that. The aim of the extension is to expose librdkafka to PHP scripts, without trying to be smarter or to expose higher level abstractions (those can be done in PHP on top of the extension). This makes it easier to design php-rdkafka, and easier for applications or library to build on top of it.

IIRC the original issue is #320. I think that it's not a real issue: As stated [in rdkafka.h](https://github.com/edenhill/librdkafka/blob/18882875af641dcf22ea58ddfabfd76007769533/src/rdkafka.h#L1192-L1202), a rd_kafka_message_t serves two purposes: provide the application with a consumed message when err == 0, and report per-topic+partition consumer errors when err is 0. This also states that application must check err to decide what action to take.

So, it's the application's responsibility to decide whether `message->partition` makes sense for its use-case, depending on the value of message->err.

Issue #329, opened 2019-12-31
Throw \InvalidArgumentException "Out of range" for partition in ConsumerTopic::offsetStore()?


dirx:
Why not checking for correct partition range first (and throwing \InvalidArgumentException if out of range) before calling librdkafka like in other methods (::produce, etc.)?

nick-zh:
fair point, thx @dirx i added a pr for this

dirx:
@nick-zh Thank you.

PR #331, opened 2020-01-02
add partition check

nick-zh:
#329

### DIFF
--- a/topic.c
+++ b/topic.c
@@ -447,6 +447,11 @@ PHP_METHOD(RdKafka__ConsumerTopic, offsetStore)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


Steveb-p:
I didn't check it too thoroughly, but extension will be able to work after this exception is thrown, i.e. it won't be put into an unrecoverable state?

We might want a test for it?

nick-zh:
@Steveb-p i want a lot more tests for everything but first i want to get #326 waiting, otherwise it will be a painful wait
should be fine actually, as @dirx mentioned in the issue, we do this almost everywhere, so this is more of a consistency "update" :smile:

arnaud-lb:
Thanks !

Issue #330, opened 2019-12-31
Fatal error on KafkaConsumerTopic deconstruction


dirx:
Stumbled over `Fatal error: Uncaught Exception: RdKafka\Kafka::__construct() has not been called in [no active file] on line 0` after php script ended.

Seems like strange behavior in KafkaConsumerTopic deconstruction. Works for ConsumerTopic & ProducerTopic.

You may re-try with (additionally unsetting $topic explicitly):

```php
<?php

declare(strict_types=1);

use RdKafka\Conf;
use RdKafka\KafkaConsumer;

$conf = new Conf();
$conf->set('group.id','test');
$consumer = new KafkaConsumer($conf);
$topic = $consumer->newTopic('test');
//sleep(2);
//unset($topic);
```

 * PHP version: 7.4.1
 * librdkafka version: 1.3.0
 * php-rdkafka version: 4.0.3-dev & 3.1.3
 * kafka version:

Same on:

 * PHP version: 7.3.13
 * librdkafka version: 1.3.0
 * php-rdkafka version: 3.1.3
 * kafka version:

nick-zh:
@dirx another good find, thx :+1: 
i am almost certain we actually shouldn't use this, because with the high level consumer you don't need to care about the topic object :smile:

nick-zh:
Correction, we need it for metadata, I'll have another look today

dirx:
@nick-zh My ffi based librdkafka binding tests did run successfully against latest rdkafka extension 4.0.4-dev on this issue (see https://travis-ci.org/dirx/php-ffi-librdkafka/jobs/649612863#L1472).

Thank you for fixing this.

PR #332, opened 2020-01-02
remove high level consumer topic

nick-zh:
#330


nick-zh:
@arnaud-lb i am pretty sure we actually don't need this, let me know if i am wrong

nick-zh:
I'll fix this another way, we need this for metadata

PR #333, opened 2020-01-03
fix topic deconstruct for high level

nick-zh:
#330

### DIFF
--- a/topic.c
+++ b/topic.c
@@ -47,7 +47,7 @@ static void kafka_topic_free(zend_object *object TSRMLS_DC) /* {{{ */
 {
     kafka_topic_object *intern = get_custom_object(kafka_topic_object, object);
 
-    if (intern->zrk) {
+    if (ZE_ISDEF(intern->zrk)) {
         kafka_object *kafka_intern = get_custom_object_zval(kafka_object, P_ZEVAL(intern->zrk));
         if (kafka_intern->rk) {
             zend_hash_index_del(&kafka_intern->topics, (zend_ulong)intern);
--- a/topic.c
+++ b/topic.c
@@ -47,7 +47,7 @@ static void kafka_topic_free(zend_object *object TSRMLS_DC) /* {{{ */
 {
     kafka_topic_object *intern = get_custom_object(kafka_topic_object, object);
 
-    if (intern->rkt) {
+    if (intern->zrk) {
         kafka_object *kafka_intern = get_custom_object_zval(kafka_object, P_ZEVAL(intern->zrk));
         if (kafka_intern->rk) {
             zend_hash_index_del(&kafka_intern->topics, (zend_ulong)intern);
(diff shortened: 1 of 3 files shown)


nick-zh:
@arnaud-lb let me know if we can do this in a better way, but since high level is different i just moved the logic to the topic, since we don't want to throw the exception in the high level case.
I think it is because we are using different structs, not sure if there is a better way to check this, let me know.

arnaud-lb:
Interesting issue. We store the zval of a `\RdKafka` object in `kafka_topic_object->zrk`. When destroying the topic, we treat `->zrk` as such.

However, when the topic has been created by a the high level consumer, `->zrk` is the zval of a `\RdKafka\KafkaConsumer`.

topic.c
```diff
@@ -48,8 +48,8 @@ static void kafka_topic_free(zend_object *object TSRMLS_DC) /* {{{ */
     kafka_topic_object *intern = get_custom_object(kafka_topic_object, object);
 
     if (intern->rkt) {
-        kafka_object *kafka_intern = get_kafka_object(P_ZEVAL(intern->zrk) TSRMLS_CC);
-        if (kafka_intern) {
+        kafka_object *kafka_intern = get_custom_object_zval(kafka_object, P_ZEVAL(intern->zrk));
```
arnaud-lb: Using `get_custom_object_zval()` prevents the exception, however we are now treating a `\RdKafka\KafkaConsumer`'s `struct _object_intern` as a `\RdKafka`'s `struct _kafka_intern` which is invalid.

Possible fix: Let `->zrk` be `NULL` when the topic is created by a `\RdKafka\KafkaConsumer`, since we don't actually need that reference in this case. Ignore `->zrk` when is `NULL` everywhere else.

We should add a test case for this.

nick-zh: @arnaud-lb i have addapted accordingly, let me know if this is what you had in mind :v:

arnaud-lb: Looks good! We should be able to retain the original line 51 and 52 though ?

We also need to make sure that we are not using `intern->zrk` somewhere else when it's not defined.

nick-zh: since `KafkaConsumerTopic` only offers `offsetStore` (i am actually unsure if we need this at all for high level) which doesn't use `intern-zrk` we should be fine.
I reverted the if back to it's original state, thx for the input :v:

arnaud-lb:
Thanks !

zxsz4084:
I also get this error 。。。

`php --ri rdkafka
rdkafka
rdkafka support => enabled
version => 4.0.3
build date => May 21 2020 16:05:53
librdkafka version (runtime) => 1.4.2
librdkafka version (build) => 1.4.2.255
`
error:
[25-May-2020 20:22:55 PRC] PHP Fatal error:  Uncaught Exception: RdKafka\Kafka::__construct() has not been called in [no active file]:0

Steveb-p:
@zsxsoft Create an issue with it and try to provide as much details as possible. It's not the first time this is reported, but currently it was not pinpointed to specific conditions that are needed for this to happen.

nick-zh:
@zxsz4084 please refer to the open issue #367 it's about the same, but as @Steveb-p said, so far, nobody was able to help me reproduce this, so i was not able to find out the origin of this error.

PR #339, opened 2020-01-20
Use auto.offset.reset earliest instead of smallest

romainneutron:
Hello,

From the Kafka doc, "smallest" does not exist. It seems it's been deprecated and replaced by "earliest"

### DIFF
--- a/tests/conf_callbacks_integration.phpt
+++ b/tests/conf_callbacks_integration.phpt
@@ -11,7 +11,7 @@ require __DIR__ . '/integration-tests-check.php';
 
 $conf = new RdKafka\Conf();
 
-$conf->set('auto.offset.reset', 'smallest');
+$conf->set('auto.offset.reset', 'earliest');
 $conf->set('metadata.broker.list', getenv('TEST_KAFKA_BROKERS'));
 $conf->set('group.id', sprintf("test_rdkafka_group_%s", uniqid()));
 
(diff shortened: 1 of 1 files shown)


romainneutron:
See also https://github.com/arnaud-lb/php-rdkafka-doc/pull/41

Steveb-p:
This value is coming from librdkafka. See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

romainneutron:
The librdkafka supports both for BC. However, it seems the "smallest" and "largest" values are not documented anymore from version 0.11.0, see:

https://kafka.apache.org/0100/documentation.html (version 0.10)
https://kafka.apache.org/0110/documentation.html (version 0.11)

Steveb-p:
> The librdkafka supports both for BC.

There is no information that I could find that says those values have been deprecated. It's true that **Kafka** uses `earliest` and `latest` in their Consumer API documentation, but **librdkafka** (which is a C binding that phprdkafka uses) is not removing their alternatives.

Even more so, you've linked Kafka documentation that is pretty old and those values are still being available.

I'd accept this PR just to be consistent with Java docs, but it's not really a deprecation.

nick-zh:
Thank you @romainneutron , i think it would be nice to keep this consistent with the kafka doc, if you want you can add a pr to the doc repo as well, if you don't have time, i will check this later :v:

romainneutron:
@nick-zh It's already done here https://github.com/arnaud-lb/php-rdkafka-doc/pull/41

nick-zh (approved):

nick-zh:
Thanks a lot!

PR #340, opened 2020-01-27
Cast LOG_DEBUG to string in sample

Steveb-p:
All configuration parameters passed to librdkafka should be strings.
Failure to do so results in:
```
TypeError : RdKafka\Conf::set() expects parameter 2 to be string, integer given
```


nick-zh (approved):

nick-zh:
Thanks @Steveb-p :v:

PR #341, opened 2020-01-28
remove examples from repo

nick-zh:
the examples were outdated and since we keep the ones from the repo in sync, i think it is best to just link to them. It is already not optimal that we need to keep code examples from readme and doc in sync
@arnaud-lb let me know if that is ok :v:


arnaud-lb (approved):

arnaud-lb:
Thanks !

PR #349, opened 2020-02-16
Fix callback integration test by enabling EOF response

Steveb-p:


nick-zh:
@Steveb-p seems the test is still somehow broken, let me know if i can help with this

PR #350, opened 2020-02-16
Fix RD_KAFKA_RESP_ERR__TIMED_OUT when there are no more messages

### LINKED ISSUE #342: I keep getting RD_KAFKA_RESP_ERR__TIMED_OUT
* PHP version: 7.0.3
 * librdkafka version: 1.3.0
 * php-rdkafka version: latest
 * kafka version: 2.11-2.1.1

I keep getting error code -185 (RD_KAFKA_RESP_ERR__TIMED_OUT) after I use `$producer->flush(100)`
The library should be working fine since `$producer->newTopic("test");` is adding the topic to kafka topics list.
For more context, i'm compiling rdkafka library and using php-rdkafka inside a docker container.
It also might be a misunderstanding of the usage of the library or a wrong version selection.
Sorry if this is not the appropriate place to ask this question but i tried everything i can with no result.

Thank you in advance

EDIT:
Changed the error constant

### LINKED ISSUE #343: Improve RD_KAFKA_RESP_ERR__TIMED_OUT to be able to differentiate cases
https://github.com/arnaud-lb/php-rdkafka/blob/bad5cbaa2810ca5339540c7e11c277362cb736b1/kafka_consumer.c#L398

Hello, 
This is a question mostly rather than an issue :blush:

I'd like to ask why the default behavior of librdkafka (return null if no message appears during the specified timeout) has been replaced by returning the `RD_KAFKA_RESP_ERR__TIMED_OUT` error code.

To my understanding having a null message (no message) after the specified timeout is not an error case. It is something expected isn't it?

Especially in cases where a small timeout is being used the application could have many `RD_KAFKA_RESP_ERR__TIMED_OUT` error code every minute. 

Could this "hide" a timeout error that could arrive from `rd_kafka_consumer_poll` pointing out actual timeout during broker communication?

Thanks :relaxed:

Steveb-p:
This fixes #343 and fixes #342. 

Allows for long-running scripts to continue consuming messages from broker, while not making user think that there is a connectivity error. In such case librdkafka will populate `err` property anyway.

This allows long-running scripts to continue uninterrupted, without falsely reporting that a timeout is reached while in reality there simply was no new message. This is important even in cases where `enabled.partition.eof` is set to `true`, as `EOF` is emitted only once and then consumer will start returning `null` when `rd_kafka_consumer_poll` is invoked, leading to errors described in fixed issues.

I believe it should not have any memory leaks by introducing those changes, but my local development environment (docker) isn't built from source and doesn't have the required debug symbols, so I'm relying on travis tests to spot them (since I'm no expert in PHP extensions).

### DIFF
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -381,7 +381,7 @@ PHP_METHOD(RdKafka__KafkaConsumer, consume)
 {
     object_intern *intern;
     zend_long timeout_ms;
-    rd_kafka_message_t *rkmessage;
+    rd_kafka_message_t *rkmessage, rkmessage_tmp = {0};
 
     if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "l", &timeout_ms) == FAILURE) {
         return;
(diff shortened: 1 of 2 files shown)


nick-zh:
@Steveb-p i might be wrong, but my point in the issue was, that if we leave the default of `enabled.partition.eof` (which is false), we cannot differentiate anymore since both poll timeout and eof will return null. So we would need to do more to differentiate those states, probably with an additional const just for the extension, this would be a breaking change though (major release), this is why i argued the value of it since most consumers will just continue polling anyway i guess.

PR #359, opened 2020-03-16
Transactional producer

### LINKED ISSUE #358: Add transactional producer
Support transactions for producers (https://github.com/edenhill/librdkafka/pull/2605)

nick-zh:
Fixes #358

### DIFF
--- a/kafka_error_exception.c
+++ b/kafka_error_exception.c
@@ -56,7 +56,7 @@ void kafka_error_new(zval *return_value, const rd_kafka_error_t *error TSRMLS_DC
 
     zend_call_method_with_2_params(return_value, ce_kafka_exception, NULL, "__construct", NULL, &message, &errorCode);
 
-    zend_update_property_string(ce_kafka_error, return_value, ZEND_STRL("string"), rd_kafka_error_string(error) TSRMLS_CC);
+    zend_update_property_string(ce_kafka_error, return_value, ZEND_STRL("error_string"), rd_kafka_error_string(error) TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, return_value, ZEND_STRL("isFatal"), rd_kafka_error_is_fatal(error) TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, return_value, ZEND_STRL("isRetriable"), rd_kafka_error_is_retriable(error) TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, return_value, ZEND_STRL("transactionRequiresAbort"), rd_kafka_error_txn_requires_abort(error) TSRMLS_CC);
@@ -71,11 +71,11 @@ object_intern * get_object(zval *zerr TSRMLS_DC) /* {{{ */
 }
 /* }}} */
 
-/* {{{ proto RdKafka\KafkaErrorException::__construct(string $message, int $code[, string $string, bool $isFatal, bool $isRetriable, bool $transactionRequiresAbort]) */
+/* {{{ proto RdKafka\KafkaErrorException::__construct(string $message, int $code[, string $error_string, bool $isFatal, bool $isRetriable, bool $transactionRequiresAbort]) */
 ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_error___construct, 0, 0, 6)
     ZEND_ARG_INFO(0, message)
     ZEND_ARG_INFO(0, code)
-    ZEND_ARG_INFO(0, string)
+    ZEND_ARG_INFO(0, error_string)
     ZEND_ARG_INFO(0, isFatal)
     ZEND_ARG_INFO(0, isRetriable)
     ZEND_ARG_INFO(0, transactionRequiresAbort)
@@ -83,13 +83,13 @@ ZEND_END_ARG_INFO()
 
 PHP_METHOD(RdKafka__KafkaErrorException, __construct)
 {
-    char *message, *string = "";
-    arglen_t message_length = 0, string_length = 0;
+    char *message, *error_string = "";
+    arglen_t message_length = 0, error_string_length = 0;
     zend_bool isFatal = 0, isRetriable = 0, transactionRequiresAbort = 0;
     zend_long code = 0;
     zval errorCode, errorMessage;
 
-    if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "sl|sbbb", &message, &message_length, &code, &string, &string_length, &isFatal, &isRetriable, &transactionRequiresAbort) == FAILURE) {
+    if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "sl|sbbb", &message, &message_length, &code, &error_string, &error_string_length, &isFatal, &isRetriable, &transactionRequiresAbort) == FAILURE) {
         return;
     }
 
@@ -97,20 +97,20 @@ PHP_METHOD(RdKafka__KafkaErrorException, __construct)
     RDKAFKA_ZVAL_STRING(&errorMessage, message);
     zend_call_method_with_2_params(getThis(), ce_kafka_exception, NULL, "__construct", NULL, &errorMessage, &errorCode);
 
-    zend_update_property_string(ce_kafka_error, getThis(), ZEND_STRL("string"), string TSRMLS_CC);
+    zend_update_property_string(ce_kafka_error, getThis(), ZEND_STRL("error_string"), error_string TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("isFatal"), isFatal TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("isRetriable"), isRetriable TSRMLS_CC);
     zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("transactionRequiresAbort"), transactionRequiresAbort TSRMLS_CC);
 }
 /* }}} */
 
-/* {{{ proto void RdKafka\KafkaErrorException::getString()
+/* {{{ proto void RdKafka\KafkaErrorException::getErrorString()
     Get name of error */
 
-ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_error_get_string, 0, 0, 0)
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_error_get_error_string, 0, 0, 0)
 ZEND_END_ARG_INFO()
 
-PHP_METHOD(RdKafka__KafkaErrorException, getString)
+PHP_METHOD(RdKafka__KafkaErrorException, getErrorString)
 {
     object_intern *intern;
     zval *res;
@@ -125,7 +125,7 @@ PHP_METHOD(RdKafka__KafkaErrorException, getString)
         return;
     }
 
-    res = rdkafka_read_property(ce_kafka_error, getThis(), ZEND_STRL("string"), 0 TSRMLS_CC);
+    res = rdkafka_read_property(ce_kafka_error, getThis(), ZEND_STRL("error_string"), 0 TSRMLS_CC);
 
 #if PHP_MAJOR_VERSION >= 7
     ZVAL_DEREF(res);
@@ -235,7 +235,7 @@ PHP_METHOD(RdKafka__KafkaErrorException, transactionRequiresAbort)
 
 static const zend_function_entry kafka_error_fe[] = { /* {{{ */
     PHP_ME(RdKafka__KafkaErrorException, __construct, arginfo_kafka_error___construct, ZEND_ACC_PUBLIC)
-    PHP_ME(RdKafka__KafkaErrorException, getString, arginfo_kafka_error_get_string, ZEND_ACC_PUBLIC)
+    PHP_ME(RdKafka__KafkaErrorException, getErrorString, arginfo_kafka_error_get_error_string, ZEND_ACC_PUBLIC)
     PHP_ME(RdKafka__KafkaErrorException, isFatal, arginfo_kafka_error_is_fatal, ZEND_ACC_PUBLIC)
     PHP_ME(RdKafka__KafkaErrorException, isRetriable, arginfo_kafka_error_is_retriable, ZEND_ACC_PUBLIC)
     PHP_ME(RdKafka__KafkaErrorException, transactionRequiresAbort, arginfo_kafka_error_transaction_requires_abort, ZEND_ACC_PUBLIC)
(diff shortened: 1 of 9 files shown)


rdkafka.c
```diff
@@ -766,8 +766,136 @@ PHP_METHOD(RdKafka__Producer, __construct)
...
+}
+/* }}} */
+
+#endif
+
 static const zend_function_entry kafka_producer_fe[] = {
     PHP_ME(RdKafka__Producer, __construct, arginfo_kafka_producer___construct, ZEND_ACC_PUBLIC)
+#ifdef HAS_RD_KAFKA_TRANSACTIONS
+    PHP_ME(RdKafka__Producer, initTransactions, arginfo_kafka_init_transactions, ZEND_ACC_PUBLIC)
+    PHP_ME(RdKafka__Producer, beginTransactionTransactions, arginfo_kafka_begin_transaction, ZEND_ACC_PUBLIC)
```
Steveb-p: Looks like a typo to me :)

nick-zh: today is typo day :v:

nick-zh:
After short discussion with @arnaud-lb, i agree that it is probably better to expose `rd_kafka_error_t` as `Rdkafka\KafkaError` to give the user even more control and transparency how to react to an error.
I will update accordingly in the coming days.

kafka_error.c
```diff
@@ -0,0 +1,224 @@
...
+
+PHP_METHOD(RdKafka__KafkaError, getName)
+{
+    object_intern *intern;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "") == FAILURE) {
+        return;
+    }
+
+    intern = get_object(getThis() TSRMLS_CC);
```
arnaud-lb: `intern` may be null here. We need to return in this case, otherwise it will seg fault :)

arnaud-lb: In `kafka_consumer.c`, for example, we do this:

```
       intern = get_object(getThis() TSRMLS_CC);                                   
       if (!intern) {                                                              
           return;                                                                 
       }                                                                           
```

kafka_error.c
```diff
@@ -0,0 +1,224 @@
...
+object_intern * get_object(zval *zerr TSRMLS_DC) /* {{{ */
+{
+    object_intern *oerr = get_custom_object_zval(object_intern, zerr);
+
+    return oerr;
+}
+/* }}} */
+
+/* {{{ private constructor */
+PHP_METHOD(RdKafka__KafkaError, __construct)
```
arnaud-lb: I _think_ that this constructor is not used (it's not in the `zend_function_entry` bellow). See `topic.c` for how to declare it as private.

About it being private: It would be useful to make it public, as this would make it possible to instantiate it in PHP tests, for example. I found myself instantiating the `Message` class in some of my tests, for example. We should not do that for behavior objects like `Producer`, but for value objects like `Message` or `KafkaError`, it can be very useful.

Maybe we should leave it like this, but then expose static constructors that mimic the `rd_kafka_error_new_fatal`, `rd_kafka_error_new_retriable`, `rd_kafka_error_new_txn_requires_abort` function (from rdkafka_error.h). This would allow to do this:

```
$error = KafkaError::createRetriable(...);
```

kafka_error.c
```diff
@@ -0,0 +1,224 @@
...
+}
+/* }}} */
+
+/* {{{ proto void RdKafka\KafkaError::getMessage()
+    Get message of error */
+
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_error_get_message, 0, 0, 0)
+ZEND_END_ARG_INFO()
+
+PHP_METHOD(RdKafka__KafkaError, getMessage)
```
arnaud-lb: `toString()` ?

kafka_error.c
```diff
@@ -0,0 +1,224 @@
...
+  +----------------------------------------------------------------------+
+  | Author: Arnaud Le Blanc <arnaud.lb@gmail.com>                        |
+  +----------------------------------------------------------------------+
+*/
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#ifdef HAS_RD_KAFKA_TRANSACTIONS
```
arnaud-lb: We could also exclude this file from the build, as we did for `kafka_consumer.c`

rdkafka.c
```diff
@@ -766,8 +770,143 @@ PHP_METHOD(RdKafka__Producer, __construct)
...
+        return;
+    }
+
+    error = rd_kafka_init_transactions(intern->rk, timeout_ms);
+
+    if (NULL == error) {
+        RETURN_LONG(RD_KAFKA_RESP_ERR_NO_ERROR);
+    }
+
+    kafka_error_new(return_value, error TSRMLS_CC);
```
arnaud-lb: What do you think about making KafkaError extend \Exception, and throwing a KafkaError here (and returning null above) ?

arnaud-lb: I think that this would be consistent with other methods. For example, `KafkaConsumer::assign()` throws an exception in case of error.

tests/produce_consume.phpt
```diff
@@ -48,6 +57,10 @@ while ($producer->getOutQLen()) {
     $producer->poll(50);
 }
 
+if (class_exists('RdKafka\KafkaError')) {
+    $producer->commitTransaction(10000);
+}
```
arnaud-lb: I'm worried that when `RdKafka\KafkaError` exists, we are not testing the non-transactional producer anymore :)

kafka_error.c
arnaud-lb: (or maybe `::newRetriable()`, to be consistent with `newQueue()`, `newTopic()`)

kafka_error.c
arnaud-lb: (as the native function is call `_string()`)

rdkafka.c
arnaud-lb: We could even extend `\Rdkafka\Exception`

kafka_error.c
nick-zh: I totally agree :v: thanks for the input

kafka_error.c
nick-zh: as string is not really telling us what it is, i prefer something more meaningful like getMessage or getReason :smile:

rdkafka.c
nick-zh: I will re-evaluate, it felt a bit wrong to have so many getters on an exception, but i agree, overall it is probably more consistent.

tests/produce_consume.phpt
nick-zh: totally true, i missed that, thx

kafka_error.c
arnaud-lb: I agree, but it makes people's life easier if we follow librdkafka's naming as much as possible :)

kafka_error.c
nick-zh: @arnaud-lb can you maybe elaborate? I tried to find how this is done for `kafka_consumer.c` but somehow did not find it :sweat_smile:

rdkafka.c
nick-zh: @arnaud-lb What do you think about naming it KafkaTransactionException, so far it is only used for this. Naming it KafkaErrorException feels a bit silly :smile:

kafka_error.c
arnaud-lb: Search for kafka_consumer.c in config.m4 :) We add the file in the list of sources, when a AC_CHECK_LIB() is successful.

kafka_error.c
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    intern = get_custom_object_zval(object_intern, getThis());
+
+    zend_update_property_string(ce_kafka_error, getThis(), ZEND_STRL("message"), message TSRMLS_CC);
+    zend_update_property_long(ce_kafka_error, getThis(), ZEND_STRL("code"), code TSRMLS_CC);
+    zend_update_property_string(ce_kafka_error, getThis(), ZEND_STRL("string"), string TSRMLS_CC);
+    zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("isFatal"), isFatal TSRMLS_CC);
+    zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("isRetriable"), code TSRMLS_CC);
+    zend_update_property_bool(ce_kafka_error, getThis(), ZEND_STRL("transactionRequiresAbort"), code TSRMLS_CC);
```
arnaud-lb: code or transactionRequiresAbort ?

kafka_error.c
```diff
@@ -0,0 +1,237 @@
...
+    ZEND_ARG_INFO(0, transactionRequiresAbort)
+ZEND_END_ARG_INFO()
+
+PHP_METHOD(RdKafka__KafkaErrorException, __construct)
+{
+    char *message, *string = "";
+    arglen_t message_length, string_length;
+    zend_bool isFatal = 0, isRetriable = 0, transactionRequiresAbort = 0;
+    zend_long code;
+    object_intern *intern;
```
arnaud-lb: I think that some of these are used uninitialized

kafka_error.c
```diff
@@ -0,0 +1,237 @@
...
+    set_object_handler_free_obj(&handlers, kafka_error_free);
+    set_object_handler_offset(&handlers, XtOffsetOf(object_intern, std));
+
+    INIT_NS_CLASS_ENTRY(ce, "RdKafka", "KafkaErrorException", kafka_error_fe);
+    ce_kafka_error = rdkafka_register_internal_class_ex(&ce, ce_kafka_exception TSRMLS_CC);
+
+	zend_declare_property_null(ce_kafka_error, ZEND_STRL("string"), ZEND_ACC_PUBLIC TSRMLS_CC);
+    zend_declare_property_bool(ce_kafka_error, ZEND_STRL("isFatal"), 0, ZEND_ACC_PUBLIC TSRMLS_CC);
+    zend_declare_property_bool(ce_kafka_error, ZEND_STRL("isRetriable"), 0, ZEND_ACC_PUBLIC TSRMLS_CC);
+    zend_declare_property_bool(ce_kafka_error, ZEND_STRL("transactionRequiresAbort"), 0, ZEND_ACC_PUBLIC TSRMLS_CC);
```
arnaud-lb: I think that some of them are missing.
Also, we should declare them as private.

arnaud-lb: As an alternative, we could store the `rd_kafka_error_t` in the `_object_intern` struct. This way, we would not need properties at all, which would simplify the code of this class. This can work if __construct is private (and does nothing, or maybe it just calls the parent constructor), plus if we add named constructors like `::newRetriable()`, `::newFatal()`, and so on. In these methods we could just have to call the corresponding `rd_kafka_error_new_xxx()` method to create a `rd_kafka_error_t`, and create a new instance of KafkaErrorException with that.

nick-zh: When i declared them as private i had an error, i will recheck what it was.
I think, as you mentioned before, for testing reasons, probably having the class like this, makes it a bit easier. Also, but i am not sure what best practice is, i think if we can avoid storing an object (and freeing it again) maybe that's better / faster.

kafka_error.c
nick-zh: good catch, sry :sweat_smile:

kafka_error.c
nick-zh: Changed to private and fixed naming inconsistencies, seems to work fine :v: , thx again

kafka_error.c
nick-zh: i think we removed that when we dropped support for anything below `librdkafka:0.11.x` i added this as suggested, thx for this input :+1:

nick-zh:
@arnaud-lb implemented the feedback and fixed / added tests, we need to merge #372 before, so tests for `librdkafka:master` will pass again

kafka_error_exception.c
```diff
@@ -0,0 +1,264 @@
...
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "sl|sbbb", &message, &message_length, &code, &string, &string_length, &isFatal, &isRetriable, &transactionRequiresAbort) == FAILURE) {
+        return;
+    }
+
+    intern = get_custom_object_zval(object_intern, getThis());
+
+    ZVAL_LONG(&errorCode, code);
+    RDKAFKA_ZVAL_STRING(&errorMessage, message);
+    zend_call_method_with_2_params(&intern, ce_kafka_exception, NULL, "__construct", NULL, &errorMessage, &errorCode);
```
arnaud-lb: `zend_call_method_with_2_params` expects a zval as first parameter (at least in PHP 7.x)

kafka_error_exception.c
```diff
@@ -0,0 +1,264 @@
...
+    zend_class_entry ce;
+
+    handlers = kafka_default_object_handlers;
+    set_object_handler_free_obj(&handlers, kafka_error_free);
+    set_object_handler_offset(&handlers, XtOffsetOf(object_intern, std));
+
+    INIT_NS_CLASS_ENTRY(ce, "RdKafka", "KafkaErrorException", kafka_error_fe);
+    ce_kafka_error = rdkafka_register_internal_class_ex(&ce, ce_kafka_exception TSRMLS_CC);
+
+	zend_declare_property_null(ce_kafka_error, ZEND_STRL("string"), ZEND_ACC_PRIVATE TSRMLS_CC);
```
arnaud-lb:
``` suggestion
    zend_declare_property_null(ce_kafka_error, ZEND_STRL("string"), ZEND_ACC_PRIVATE TSRMLS_CC);
```

kafka_error_exception.c
```diff
@@ -0,0 +1,264 @@
...
+
+object_intern * get_object(zval *zerr TSRMLS_DC) /* {{{ */
+{
+    object_intern *oerr = get_custom_object_zval(object_intern, zerr);
+
+    return oerr;
+}
+/* }}} */
+
+/* {{{ proto RdKafka\KafkaErrorException::__construct(string $message, int $code[, string $string, bool $isFatal, bool $isRetriable, bool $transactionRequiresAbort]) */
```
arnaud-lb: I would _really_ like that we copy the librdkafka API, such as `rd_kafka_error_new_fatal()` or `rd_kafka_error_new_retriable()`, instead of using this constructor :) 

We could do this:
- Add KafkaErrorException::newFatal(), KafkaErrorException::newRetriable(), etc. These methods would be static, and return a new KafkaErrorException instance. Internally, they would do something like kafka_error_new().

We could also remove the class properties, and replace them with a `rd_kafka_error_t` member on `object_intern`. See the `rd_kafka_t` member in `object_intern` in `kafka_consumer.c` for example.

- Add a `rd_kafka_error_t` in object_intern
- Free the `rd_kafka_error_t` in `kafka_error_free()`
- Change `get_object()` such that it checks that the `rd_kafka_error_t` is not null (see `get_object()` in `kafka_consumer.c` for example)
- Make the constructor private (see `topic.c` to re-use the existing no-op constructor).

rdkafka.c
```diff
@@ -766,8 +770,147 @@ PHP_METHOD(RdKafka__Producer, __construct)
...
+   Initializes transactions, needs to be done before producing and starting a transaction */
+ZEND_BEGIN_ARG_INFO_EX(arginfo_kafka_init_transactions, 0, 0, 1)
+    ZEND_ARG_INFO(0, timeout_ms)
+ZEND_END_ARG_INFO()
+
+PHP_METHOD(RdKafka__Producer, initTransactions)
+{
+    kafka_object *intern;
+    zend_long timeout_ms;
+    const rd_kafka_error_t *error;
```
arnaud-lb, line 785: This const pointer is passed as non const pointer

kafka_error_exception.c
nick-zh: @arnaud-lb yes i remember you mentioned this previously in a comment as alternative.
Can you maybe elaborate why you feel this is the better way of doing it?
I feel transforming and storing an object feels bothersome and freeing it afterwards, if not necessary. Especially since this is an exception.
Right now i don't see the benefit we would gain.

So my opinion regarding `rd_kafka_error_new_fatal`, etc., over the constructor we have the same possibilites. This would be basically the same as `rd_kafka_error_new` and `rd_kafka_error_set_fatal`, just in one go though. Also except for tests, nobody will make use of this :+1: 

Let me know what you think, of course if you insist i can still change it to be using the object.

rdkafka.c
nick-zh: awesome thx, i missed that :+1:

kafka_error_exception.c
nick-zh: adjusted, thx for the hint :+1:

arnaud-lb:
Thanks @nick-zh !

arnaud-lb:
I've released version 4.1.0 with this feature

PR #372, opened 2020-06-01
remove addBroker from tests

nick-zh:
It seems `addBroker` is not favourited anymore and brokers should be in config.
Root cause was tests failing on `librdkafka:master`.


nick-zh:
@arnaud-lb please have a look, without this our current tests fail on `librdkafka:master`

arnaud-lb:
Thanks!

PR #383, opened 2020-08-26
Add PHP8 support, drop PHP5 support

### LINKED ISSUE #328: Dropping PHP EOL versions with next major, prepare for PHP8
With `php:8.0` approaching next year, i think it would make sense to deprecate all
php versions that have reached EOL.
The versions that are or will be EOL when 8.0 comes out are:
- `php:5.6` (EOL since 12/2018)
- `php:7.0` (EOL since 12/2018)
- `php:7.1` (EOL since 12/2019)
- `php:7.2` (EOL with `php:8.0`)

I think this would help clean the code up and also reduce CI time.
It would probably make sense, to start a branch for this later next year,
to clean stuff up and make it `php:8.0` compatible.
We probably should also drop support for anything before `librdkafka:1.2.0` by then.
What are your thoughts @arnaud-lb ?

nick-zh:


nick-zh:
Need to check how i can test the segfaults, locally with 8.0.0beta2 i can't reproduce it :cry: 

**Edit:** Also built a php image from master, was still unable to reproduce

nick-zh:
Ok found it, so Travis is using a build from commit `542ef745333ee27104ce97144b99910add2843c0` or before. This version still has zval in the `zend_update_property_*` instead of zend_object. The segfaults should be resolved, once the nightly image of Travis [has been updated](https://travis-ci.community/t/update-php-nightly/9734).

tests/conf_setDefaultTopicConf.phpt
```diff
@@ -2,7 +2,7 @@
 RdKafka\Conf::setDefaultTopicConf()
 --SKIPIF--
 <?php
-if (!method_exists('RdKafka\Conf', 'setDefaultTopicConf')) {
+if (!method_exists('RdKafka\Conf', 'setDefaultTopicConf') || 7 < PHP_MAJOR_VERSION) {
```
Steveb-p, line 5: Won't the second condition be always false, since we're always testing PHP 7 & 8 now?

nick-zh: `7<8 => true`

Steveb-p: Duh, yoda-style conditions always get me :D

nick-zh:
@arnaud-lb please have a look if you can spot further improvements. Otherwise i think after we've released `php-rdkafka:4.1` we can merge this into the master, so i can get started on the other features / improvements that will be included for `php-rdkafka:5.0`

nick-zh:
@arnaud-lb regarding `zeval.h` shall we drop it completely and replace the macros and types (since 7.x only uses zend engine 3)? What is your opinion?

remicollet:
Looks like there is still a lot of work toward PHP 8

Especially all the properties management and get_debug_info using zend_object instead of zval.
This raise lot of build warnings (and obviously the pointer cast are not OK)

nick-zh:
@remicollet thx a lot for the input, i am still somewhat new to ext development and learning, so i really appreciate this. I will try to fix the points you mentioned this week :v:

nick-zh:
@remicollet i fixed some of the things you mentioned, i saw one warning regarding zval_zval, will check it out later

remicollet:
Seems far better ;)

nick-zh:
Yeah thx for the input, totally missed the warnings and was locally on an outdated php master build,
i will fix the remaining warning and have a look over again on the weekend (comparing to core and other exts if there is more i can improve)
PS: somehow i can't find rdkafka on your ext list anymore, i saw it yesterday with the link though, not sure what happened :sweat_smile:

remicollet:
> PS: somehow i can't find rdkafka on your ext list anymore, i saw it yesterday with the link though, not sure what happened sweat_smile

Indeed.... fixed (huge work to maintain such a list as an HTML page...)

nick-zh:
Thx a lot. Yeah i can imagine, but thx for it, it is really good to have an overview :tada:

pascaldevink:
@nick-zh @remicollet @arnaud-lb What is the status of this PR? Can it be merged now that PHP 8 has been released?

nick-zh:
It is up to @arnaud-lb to review it, also we have a pending `4.1` release atm that he probably will release first
You can always build the branch if you need it fast :v:

arnaud-lb:
Congrats for the huge work @nick-zh !

There is more than just droping PHP 5 support here:

* This drops ZTS support on PHP 7. I believe that it's easy to continue supporting it, so I would like to keep that (we just have to define TSRMLS_xx to no-ops on PHP 8).

* This increases the librdkafka version requirement to 1.4.0. Distros do not have this version yet, so it would make it more difficult to install or package the extension. For instance, the latest Ubuntu LTS [has only librdkafka 1.2.x](https://packages.ubuntu.com/source/focal/librdkafka). Ubuntu Xenial LTS and Debian Buster both have librdkafka 0.11.x and 3rd party PHP 8 packages. Since supporting librdkafka 0.11.x is not super painful (there is only 4 conditional compilations: message headers, purge, transactions, and a partitioner), and since it would make people's life easier, I think that we should keep 0.11.x support.

I will re-add ZTS support if you rebase the branch :)

nick-zh:
@arnaud-lb regarding ZTS support, this was on purpose, the next major shouldn't support PHP versions that are EOL, people who rely on an outdated PHP version can still use `php-rdkafka:4.x` which has a lot of features, including producer transactions once you release `php-rdkafka:4.1`
Same goes for librdkafka, the next minor `librdkafka:1.3.x` is over a year old and it is not to our advantage to support old versions in the new major. A lot of bugs that get reported are because of outdate `librdkafka` versions, also it increases our build time.
I highlighted this in the issue / discussion i opened a year ago: #328

arnaud-lb:
PHP 7.4 is supported until November 2022 (security fixes) by the PHP project, and is shipped with Ubuntu Focal which is supported until 2025.

librdkafka 1.3 or 1.4 are not shipped with any production-worth distros right now, which means that php-rdkafka users might not be able to install php-rdkafka before years if we support only 1.4.

> can still use php-rdkafka:4.x which has a lot of features

If we go this route we have two php-rdkafka version to maintain, because we will not stop fixing bugs in the 4.x if people are stuck with this version.

Software that is a few years old is not necessarily outdated software, it's stable software that won't break the prod because that's what the prod is running. This has a lot of value. People hate to be forced to upgrade everything because some package has decided to stop supporting some version. I personaly hate that, and I love software that can add new features over time without breaking everything at each version.

nick-zh:
So to clarify, as mentioned in my issue:
- support only php:7.3 or greater
- support only librdkafka:1.4 or greater

I created this issue as early as possible to inform all users and have an open discussion about it.

If you are referring to my changes regarding `TSRM`, from what i read this is not necessary anymore at least according to `UPGRADE.INTERNALS`:
```
The TSRM layer undergone significant changes. It is not necessary anymore to 
pass the tsrm_ls pointer explicitly. The TSRMLS_* macros should not be used
in any new developments.
``` 
If i broke something inadvertently, then i agree we should ensure compatibility for `php:7.3` upward.

Regarding librdkafka, it is actually really easy to install it from source. People using outdated `librdkafka` versions shipped
by the distro, is a big problem not only for them (buggy releases), but also for us, since we need to debug every time if a reported bug is caused by an outdated `librdkafka` version or by the extension itself. I see no upsides in supporting anything below `librdkafka:1.4`. As stated before, `librdkafka` is evolving quickly and a lot has changed since `0.11.x` (realeased over 2 years ago).

Moving from `php:7.x` to the new minimal required `PHP:7.3` should be reasonable effort and we should encourage people
doing it.

This is your extension, so if you chose to re-add compatibilites that i dropped, it is your right to do so.
Again people are not forced to upgrade, i am totally aware that i would need to provide bug fix support for `php-rdkafka:4.x`,
but i think this would be a worthy price. We did the same back then for `php-rdkafka:3.x`.

arnaud-lb:
> I created this issue as early as possible to inform all users and have an open discussion about it.

I have some responsibility here, I should have reacted earlier.

> If you are referring to my changes regarding TSRM, from what i read this is not necessary anymore at least according to UPGRADE.INTERNALS

Yes, the TSRMLS_xx macros are not necessary anymore in PHP 8. However, PHP 7 will not build in ZTS mode without them. We can easily support both by defining them again in PHP 8:

```
#if PHP_MAJOR_VERSION >= 8
#define TSRMLS_D
#define TSRMLS_C
#define TSRMLS_DC
#define TSRMLS_CC
#endif
```

and then continuing to use them in function declarations and function calls. I can revert the removal of them in your branch if you rebase it.

> Regarding librdkafka, it is actually really easy to install it from source

If you are a package maintainer, you also need to create a librdkafka package because it's a dependency of you php-rdkafka package.
If you are provisioning some server, that's also more work than just ensuring the librdkafka package is installed.
You might also need to handle security updates, something that is usually handled by your distro.

I think that we should continue to support PHP 7.x (at least the supported versions) in ZTS and non-ZTS mode, and librdkafka 0.11.

nick-zh:
The quote from `UPGRADE.INTERNALS` is from the `php:7.x` branch, but i am not too familiar with the ZTS details, but from what i can see they are also already empty defines in `php:7.x`.
Regarding `librdkafka`, i still feel it is fair enough to drop support for those old versions for `php-rdkafka:5.x`, that's imho what majors are for. I purposefully added producer transcations already for `php-rdkafka:4.x`, since this is one of the bigger features.
`php-rdkafka:4.x` is very feature rich and supports: 
- `php:5.x` or higher
- `librdkafka:0.11.x` or higher

So nobody is forced to move to `php-rdkafka:5.x` unless they want to use `php:8`.

I personally would drop old baggage for the new major :smile: 
I will rebase my branch shortly, so you can adjust as you see fit :v:

arnaud-lb:
> The quote from UPGRADE.INTERNALS is from the php:7.x branch, but i am not too familiar with the ZTS details, but from what i can see they are also already empty defines in php:7.x.

Oh well, you are right! I knew there had been some work on this in 7.x, but not that the defines were nooped entirely. 

Thank you for the rebase!

nick-zh:
I think i messed something up with the rebase, i see missing changes, will try to resolve

nick-zh:
Should be ok again, sry for the mixup, i usually don't dabble with rebase :smile:

arnaud-lb:
I re-added support for librdkafka 0.11.

I also moved the test suite to Github Actions and made some optimizations so that we could continue to test against older librdkafka versions without increasing the build time too much. It turns out that the full test suite, including the valgrind one, now runs in under 4 minutes :tada:

andypost:
It looks great 👍 gonna check it asap for Alpinelinux

andypost:
Btw looking on memcache extension new major version looks great for 8.0

andypost:
@arnaud-lb Thamk you! it passed all offiline tests on a set of arches https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/15458 :rocket: 

Using **librdkafka (1.5.2)**

andypost (approved):

arnaud-lb:
This has been merged as a series of smaller PRs: #401, #403, #405, and #406 (author was retained in commits and PRs).

Summary of changes:
- Dropped PHP 5 support
- Added PHP 8 support

librdkafka versions that were supported are still supported. PHP 7.3 and 7.4 are still supported (7.0, 7.1, 7.2 may still work).

Thanks @nick-zh

andypost:
Awesome! Waiting for release, thank you

Jeroeny:
Is the Windows DLL that's build per version also managed by this repo? It usually appears at https://pecl.php.net/package/rdkafka. Or is that being done somewhere else? (For 5.0.0 in this case).

Edit: possible related: https://github.com/arnaud-lb/php-rdkafka/issues/440

BehzadV:
Sorry, but is the version 5 getting a windows release? Our Linux/Windows mixed team is running into troubles over this. The previous versions don't work on PHP 8.0.

nick-zh:
I will try to check #440 this weekend, i hope i can find some time. Atm i am really busy. Apologies for the delay on this

PR #392, opened 2020-09-18
Add debug to issue template

Steveb-p:
People often have issues that can be resolved only when looking at php-rdkafka logs with librdkafka in high verbosity mode.  
This adds the relevant commands to issue template so that more people start by looking at them, possibly auto-resolving their problem.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -19,6 +19,4 @@ $conf = new RdKafka\Conf();
 // Add the two settings below for a more verbose output
 $conf->set('log_level', (string) LOG_DEBUG);
 $conf->set('debug', 'all');
-$rk = new RdKafka\Producer($conf);
-$rk->addBrokers("10.0.0.1:9092,10.0.0.2:9092");
 ```
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -11,3 +11,13 @@ Please always include the following in an issue, so we can help faster:
  * librdkafka version:
  * php-rdkafka version:
  * kafka version:
+
+If possible, add debug logs from librdkafka, which you can obtain by adjusting your configuration:
+```php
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE
```diff
@@ -11,3 +11,13 @@ Please always include the following in an issue, so we can help faster:
...
  * kafka version:
+
+If possible, add debug logs from librdkafka, which you can obtain by adjusting your configuration:
+```php
+<?php
+$conf = new RdKafka\Conf();
+$conf->set('log_level', (string) LOG_DEBUG);
+$conf->set('debug', 'all');
+$rk = new RdKafka\Producer($conf);
+$rk->addBrokers("10.0.0.1:9092,10.0.0.2:9092");
```
nick-zh: maybe omit those, since they are not part of what we need :D 
```suggestion
```

.github/ISSUE_TEMPLATE
```diff
@@ -11,3 +11,13 @@ Please always include the following in an issue, so we can help faster:
  * librdkafka version:
  * php-rdkafka version:
  * kafka version:
+
+If possible, add debug logs from librdkafka, which you can obtain by adjusting your configuration:
+```php
+<?php
+$conf = new RdKafka\Conf();
+$conf->set('log_level', (string) LOG_DEBUG);
```
nick-zh, line 20:
```suggestion
// Add the two settings below for a more verbose output
$conf->set('log_level', (string) LOG_DEBUG);
```

.github/ISSUE_TEMPLATE
Steveb-p: I was basically going out of the office and just wanted to not forget to adjust the issue template, as it often happens with ideas :)

nick-zh (approved):

nick-zh:
Thx @Steveb-p

PR #393, opened 2020-09-22
Fix test

remicollet:
To avoid failure when no server configured


nick-zh:
@remicollet thank you :v:

PR #404, opened 2020-12-07
Fix pecl build (windows)

### LINKED ISSUE #402: php-rdkafka 4.1.0 requires librdkafka >= 1.4.0
I've noticed that the PECL builds for php-rdkafka 4.1.0 were failing, because `rd_kafka_error_t` is not defined. Apparently, 4.1.0 requires at least librdkafka 1.4.0, what is not documented in the [changelog](https://pecl.php.net/package-changelog.php?package=rdkafka&release=4.1.0). The [setup documentation](https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.requirements.html) even says that librdkafka >= 0.11.x would be supported.

nick-zh:
in `config.m4`  we exclude `kafka_error_exception.c` for older `librdkafka` versions, not so in the `config.w32`
I added a block around the c class as well to be sure that we don't get errors in combination with old lib versions

### DIFF
--- a/kafka_error_exception.c
+++ b/kafka_error_exception.c
@@ -20,6 +20,7 @@
 #include "config.h"
 #endif
 
+#ifdef HAS_RD_KAFKA_TRANSACTIONS
 #include "php.h"
 #include "php_rdkafka.h"
 #include "php_rdkafka_priv.h"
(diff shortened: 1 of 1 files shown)


nick-zh:
@arnaud-lb it was confirmed by the reporter, that this will fix the issue, when you have time, you can merge and release `php-rdkafka:4.1.1` from the `4.1.x` branch. Thx in advance :v:

arnaud-lb:
Thanks @nick-zh

PR #414, opened 2020-12-21
AppVeyor CI integration

cmb69:
For now we do the build only.


nick-zh:
@arnaud-lb maybe in a later stage, we can convert this into a GHA, if i am not mistaken we should have a win platform there as well

arnaud-lb:
Thanks @cmb69 !


@nick-zh +1

PR #415, opened 2020-12-21
uint is not universally available

cmb69:
On Windows and macOS, `uint` is not (necessarily) available, so we use
the explicit `unsigned int` instead.

---

To clarify: PHP had a compatibility definition of `uint` prior to 7.4.0, but that has been removed.

Furthermore, this is a straight forward replacement. It might be more correct to use `size_t` instead, since `rd_kafka_header_cnt()` returns `size_t` and `rd_kafka_header_get_all()` expects `size_t` at least with librdkafka 1.5.3.


arnaud-lb:
Thanks @cmb69 !

Issue #390, opened 2020-09-18
Feature request: support converting error code to enum string using rd_kafka_err2name


james-johnston-thumbtack:
It would be nice if php-rdkafka could expose the `rd_kafka_err2name` function, similar to how it already does with `rd_kafka_err2str`.  The former returns the name of the enum constant (e.g. `RD_KAFKA_RESP_ERR__TIMED_OUT`), whereas the latter returns a human-readable error string.  Getting the enum constant would be useful for some tracking/logging situations where the constant name is easier to work with than the raw integer value.

It may be the case that it is better to do your own implementation instead of actually calling `rd_kafka_err2name`, if the PHP constant names exposed by php-rdkafka don't exactly match the C language constant names exposed by librdkafka.  For example, if https://github.com/arnaud-lb/php-rdkafka/blob/a0c04cbb4a8b08db9afebd0748cd2c377d091ef9/rdkafka.c#L803 retained a mapping of error code to string for later use by the proposed `rd_kafka_err2name` function exposed by php-rdkafka.

nick-zh:
@james-johnston-thumbtack thanks for the suggestion. Should be easy to do, will try to implement this soon

arnaud-lb:
Implemented in master

PR #436, opened 2021-01-14
Add rd_kafka_err2name

arnaud-lb:


arnaud-lb:
Related: https://github.com/arnaud-lb/php-rdkafka/issues/390

PR #439, opened 2021-01-14
Add a failing test case for integer and null header values

andrewmy:
No idea how to fix this yet.﻿


arnaud-lb:
Hi @andrewmy 

I'm not sure that Kafka knows about the type of header values: I believe that header values are just byte arrays in Kafka

PR #509, opened 2022-01-18
Always initialize Message::$headers

arnaud-lb:


healerz:
hi @arnaud-lb, do you already know when you will release this fix?

arnaud-lb:
Hi @healerz 

I've just released version 6.0.1 with this fix

healerz:
@arnaud-lb Thanks a lot

PR #540, opened 2023-06-06
Fix kafka consumer not destroyed in close

ikeberlein:
Not destroying underlying kafka consumer in close() method leads to unclean librdkafka shutdown and very often to segfaults.

### DIFF
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -536,6 +536,7 @@ PHP_METHOD(RdKafka_KafkaConsumer, close)
     }
 
     rd_kafka_consumer_close(intern->rk);
+    rd_kafka_destroy(intern->rk);
     intern->rk = NULL;
 }
 /* }}} */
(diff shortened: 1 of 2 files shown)


arnaud-lb:
Normally the `rk` handle is destroyed when the `KafkaConsumer` object itself is destroyed, so I don't think we need to close it explicitly here. However maybe we don't destroy various resources in the right order when the `rk` is closed later?

Do you have a reproducer?

plehatron:
Hey @arnaud-lb, I prepared and reported the reproducer in https://github.com/arnaud-lb/php-rdkafka/issues/571. The issue includes details about the bug this PR fixes.

mattiabasone:
Hey! Do you happen to have any news on this @arnaud-lb?
I'm experiencing the same issue during PHPUnit tests. In the CI, I'm using Ubuntu 24.04 and the shivammathur/setup-php GitHub action for setting up the environment with PHP 8.3.25 and rdkafka 6.0.5

takeit:
@arnaud-lb same issue here, could this be merged, pls? it's been over 2 years for 1 line to be merged

ralphschindler:
@ikeberlein can you look at the test in #595 - does this cover the workflow you're seeing produce a segfault?

PR #541, opened 2023-06-06
Add missing methods to support incremental rebalance

ikeberlein:
`::incrementalAssign()` and `::incrementalUnassign()` consumer methods are used in rebalance callback instead of `::assign()` when cooperative-sticky partition assignment strategy in charge.

### DIFF
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -212,6 +212,59 @@ PHP_METHOD(RdKafka_KafkaConsumer, assign)
 }
 /* }}} */
 
+#ifdef HAS_RD_KAFKA_INCREMENTAL_ASSIGN
+static void consumer_incremental_op(int assign, INTERNAL_FUNCTION_PARAMETERS) /* {{{ */
+{
+    HashTable *htopars = NULL;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "h", &htopars) == FAILURE || !htopars) {
+        return;
+    }
+
+    object_intern *intern = get_object(getThis());
+    if (!intern) {
+        return;
+    }
+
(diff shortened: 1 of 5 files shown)


kafka_consumer.c
```diff
@@ -212,6 +212,67 @@ PHP_METHOD(RdKafka_KafkaConsumer, assign)
...
+    }
+}
+/* }}} */
+#endif
+
+/* {{{ proto void RdKafka\KafkaConsumer::incrementalAssign(array $topics)
+    Incremental assignment of partitions to consume */
+PHP_METHOD(RdKafka_KafkaConsumer, incrementalAssign)
+{
+#if RD_KAFKA_VERSION >= 0x010600ff
```
arnaud-lb: Please change this so that the methods do not exist if the librdkafka version is not compatible. The stub files support `#if`/`#ifdef` as well.

Also, please use a feature test macro so that you do not need to rely on the version number (see [here](https://github.com/arnaud-lb/php-rdkafka/blob/bcd5004f461d1d3a5f879bb21280bdde6f6800c2/config.m4#L74-L79) and [here](https://github.com/arnaud-lb/php-rdkafka/blob/bcd5004f461d1d3a5f879bb21280bdde6f6800c2/rdkafka.c#L759)).

arnaud-lb (commented):
Thank you! Sorry for the late review. This looks good to me appart from one comment.

kafka_consumer.c
ikeberlein: Hi Arnaud and thank you too :)
did as you said

arnaud-lb:
Thank you!

bazilmarkov:
Hi, are there any plans to release this feature? Also, am I correct that the latest version released does not support the `cooperative-sticky` assignment strategy?

PR #546, opened 2024-01-03
feat: implement oauthbearer token refresh cb setter

cb-freddysart:


jcunhafonte (approved):
Great PR! I have been waiting for this PR for a while now. ❤️

conf.c
```diff
@@ -698,6 +734,43 @@ PHP_METHOD(RdKafka_Conf, setLogCb)
...
+    }
+
+    conf = get_kafka_conf_object(getThis());
+    if (!conf) {
+        return;
+    }
+
+    Z_ADDREF_P(&fci.function_name);
+
+    if (conf->cbs.log) {
```
arnaud-lb: Should be `conf->cbs.oauthbearer_token_refresh`

conf.c
```diff
@@ -698,6 +734,43 @@ PHP_METHOD(RdKafka_Conf, setLogCb)
...
 /* }}} */
 
+/* {{{ proto void RdKafka\Conf::setOauthbearerTokenRefreshCb(mixed $callback)
+   Set token refresh callback for OAUTHBEARER sasl */
+PHP_METHOD(RdKafka_Conf, setOauthbearerTokenRefreshCb)
+{
+#ifndef HAS_RD_KAFKA_OAUTHBEARER_TOKEN_REFRESH_CB
+    zend_throw_exception_ex(NULL, 0, "This version of rdkafka does not support the OAUTHBEARER sasl mechanism");
+    return;
+#endif
```
arnaud-lb: Please change this so that the `setOauthbearerTokenRefreshCb` method does not exist if `HAS_RD_KAFKA_OAUTHBEARER_TOKEN_REFRESH_CB` is not defined.

(stub files also support #ifdef)

arnaud-lb (commented):
Thank you! LGTM appart from a few comments

cb-freddysart:
Thank you for the speedy review @arnaud-lb ! I'm a bit new to PHP extensions but I believe I've addressed your current round of feedback.

Also, sorry for not providing a summary. You beat me to the punch. 

My goal here is to add support for `OAUTHBEARER` in such a way that we will be able to connect to AWS MSK with this library and I believe this PR is only half of that story.

The other half is adding support for `rd_kafka_oauthbearer_set_token` and `rd_kafka_oauthbearer_set_token_failure` (see https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka.h#L2207-L2209). Would you be open to adding these methods as well? I would either open a separate PR or add them here and reorganize the commits to make review easier. Whichever works for you.

arnaud-lb:
Nice work, thank you!

I'm generally open to adding any method provided by librdkafka. Please do so in a separate PR, as this will simplify reviews (maybe just one PR for the two methods, as they are very related).

Rastusik:
hey @arnaud-lb is there any plan to integrate the oauthbearer token refresh also in the high level consumer/producer?

arnaud-lb:
Hey @Rastusik 

I do not plan to implement this, but I would gladly accept contributions.

Rastusik:
Thanks for the info. I think someone is already considering working on it https://github.com/arnaud-lb/php-rdkafka/issues/580 ... I can think about the implementation if I will be forced to use IAM in MSK, but I think the easier way for now would be to just use PLAIN mechanism.

bdurkovic:
> I think someone is already considering working on it

I'd consider it gladly, but C isn't part of my toolkit unfortunately. However knowing what I know now, I'd tag the ticket as a "feature" rather than a "bug".

Rastusik:
@bdurkovic I'm already working on it https://github.com/arnaud-lb/php-rdkafka/compare/6.x...Rastusik:php-rdkafka:oauthbearer_hlc ... but it still needs some polishing

PR #547, opened 2024-01-10
feat: finalize OAUTHBEARER support

cb-freddysart:
This PR adds support for `rd_kafka_oauthbearer_set_token` and `rd_kafka_oauthbearer_set_token_failure`. See https://github.com/confluentinc/librdkafka/pull/2189/files#diff-eef17694b5807cd63a95c3d86c39752b62f8bdb46e46c59231bf02353a5daef5 for where this functionality was added to librdkafka.

With these changes, we are now able to successfully authenticate to AWS MSK via IAM. 🎉 

Full working example:

```
<?php

use Aws\Credentials\CredentialProvider;
use Aws\Signature\SignatureV4;
use GuzzleHttp\Psr7\Request;
use Psr\Http\Message\RequestInterface;

require 'vendor/autoload.php';

class AuthToken {
    public function __construct(public string $tokenValue, public int $expiresAt) {}
}

class MSKAuthTokenProvider {

    private const ENDPOINT_URL_TEMPLATE = 'https://kafka.%s.amazonaws.com';
    private const DEFAULT_TOKEN_EXPIRY_SECONDS = 900;
    private const ACTION_TYPE = "Action";
    private const ACTION_NAME = "kafka-cluster:Connect";
    private const SIGNING_NAME = "kafka-cluster";
    private const USER_AGENT_KEY = "User-Agent";
    private const VERSION_KEY = 'Version';

    public function generateAuthToken(): AuthToken
    {
        $endpoint_url = sprintf(self::ENDPOINT_URL_TEMPLATE, getenv('AWS_REGION'))
            . '?' . http_build_query([self::ACTION_TYPE => self::ACTION_NAME]);

        $request = new Request('GET', $endpoint_url);
        $credentials = CredentialProvider::defaultProvider()()->wait();
        $sig_v4_signer = new SignatureV4(self::SIGNING_NAME, getenv('AWS_REGION'));
        $expiresAt = new \DateTime('+' . self::DEFAULT_TOKEN_EXPIRY_SECONDS . ' seconds');

        $presigned = $sig_v4_signer->presign($request, $credentials, $expiresAt);
        $presigned_uri = $presigned->getUri();
        // Append User-Agent and Version **after** signing
        $presigned_uri .= '&'. self::USER_AGENT_KEY . '='.urlencode('msk-iam-php').
            '&'. self::VERSION_KEY . '='. urlencode('2020_10_22'); // Version is always static

        $base64_bytes = base64_encode($presigned_uri);
        $base64_encoded_signed_url = rtrim($base64_bytes, '=');

        return new AuthToken($base64_encoded_signed_url, $this->getExpirationTimeMs($presigned));
    }

    private function getExpirationTimeMs(RequestInterface $request): int
    {
        parse_str($request->getUri()->getQuery(), $ret);
        $signing_time = DateTimeImmutable::createFromFormat('Ymd\THis\Z', $ret['X-Amz-Date'], new DateTimeZone('UTC'))
            ->getTimestamp();
        $lifetime_seconds = $ret['X-Amz-Expires'];
        $expiration_ts_seconds = $signing_time + $lifetime_seconds;

        return $expiration_ts_seconds * 1000;
    }
}

$provider = new MSKAuthTokenProvider();
$token = $provider->generateAuthToken();

$conf = new RdKafka\Conf();
$conf->set('metadata.broker.list', 'broker list');
$conf->set('security.protocol', 'SASL_SSL');
$conf->set('sasl.mechanisms', 'OAUTHBEARER');

$conf->setOauthbearerTokenRefreshCb(function() use ($provider, $producer) {
    $token = $provider->generateAuthToken();
    $producer->oauthbearerSetToken($token->tokenValue, $token->expiresAt, 'kafka-cluster', ['one' => 'two']);
});

// Set initial token
$producer->oauthbearerSetToken($token->tokenValue, $token->expiresAt, 'kafka-cluster', ['one' => 'two']);

$topic = $producer->newTopic("test");
for ($i = 0; $i < 1; $i++) {
    $topic->produce(RD_KAFKA_PARTITION_UA, 0, "Hello from PHP $i");
    $producer->poll(0);
}

for ($flushRetries = 0; $flushRetries < 10; $flushRetries++) {
    $result = $producer->flush(10000);
    if (RD_KAFKA_RESP_ERR_NO_ERROR === $result) {
        break;
    }
}

echo "Done\n";

```


rdkafka_arginfo.h
```diff
@@ -1,87 +1,166 @@
 /* This is a generated file, edit the .stub.php file instead.
- * Stub hash: fbfdb28740208d5f909e9db261bea0aa26bfd471 */
+ * Stub hash: e469d659a320102b1d64c048e9457bb9b64d8e5d */
 
 ZEND_BEGIN_ARG_INFO_EX(arginfo_class_RdKafka___construct, 0, 0, 0)
 ZEND_END_ARG_INFO()
 
+#if (PHP_VERSION_ID >= 80100)
```
cb-freddysart: These were all added when running `gen-stub`. lmk if you'd like me to clean up everything other than the oauth related changes @arnaud-lb .

PyaeSoneAungRgn:
yes, we need this feature 🔥

btw, i already created [aws-msk-iam-sasl-signer-php](https://github.com/PyaeSoneAungRgn/aws-msk-iam-sasl-signer-php) package for Amazon MSK IAM Authentication

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+    }    
+
+    /* rd_kafka_oauthbearer_set_token(rd_kafka_t *rk,
+                               const char *token_value,
+                               int64_t md_lifetime_ms,
+                               const char *md_principal_name,
+                               const char **extensions,
+                               size_t extension_size,
+                               char *errstr,
+                               size_t errstr_size */
```
arnaud-lb: Please remove this comment

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+        extensions = safe_emalloc((extension_size * 2), sizeof(char *), 0);
+
+        int pos = 0;
+        zend_ulong num_key;
+        zend_string *extension_key_str;
+        zval *extension_zval;
+        ZEND_HASH_FOREACH_KEY_VAL(Z_ARRVAL_P(extensions_arg), num_key, extension_key_str, extension_zval) {
+            if (!extension_key_str) {
+                extension_key_str = zend_long_to_str(num_key);
+                zend_string_delref(extension_key_str);
```
arnaud-lb: `extension_key_str` has a refcount of `0` after that, which is risky. Ideally you should move this call after it's not referenced by `extensions` anymore.

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+            }
+
+            zend_string *tmp_extension_val_str;
+            zend_string *extension_val_str = zval_get_tmp_string(extension_zval, &tmp_extension_val_str);
```
arnaud-lb, line 492: `zval_get_tmp_string` allocates a new string when `extension_zval` is converted. In this case, `tmp_extension_val_str` needs to be released.

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+
+            extensions[pos++] = estrdup(ZSTR_VAL(extension_key_str));
+            extensions[pos++] = estrdup(ZSTR_VAL(extension_val_str));
```
arnaud-lb: The strings allocated by estrdup() must be freed at some point, otherwise these will leak. You can observe the leaks by running your code under valgrind while disabling the zend allocator: `USE_ZEND_ALLOC=0 valgrind php file.php`. PHP debug builds also report memory leaks by default.

Something like this would fix this issue and the two previous ones:

```
if (!extension_key_str) {
    extension_key_str = zend_long_to_str(num_key);
    extensions[pos++] = estrdup(ZSTR_VAL(extension_key_str));
    zend_string_release(extension_key_str);
} else {
    extensions[pos++] = estrdup(ZSTR_VAL(extension_key_str));
}

zend_string *tmp_extension_val_str;
zend_string *extension_val_str = zval_get_tmp_string(extension_zval, &tmp_extension_val_str);
extensions[pos++] = estrdup(ZSTR_VAL(extension_val_str));
if (tmp_extension_val_str) {
    zend_string_release(tmp_extension_val_str);
}
```

Then efree() all `extensions` element just before efree(extensions).

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+ The SASL/OAUTHBEARER token refresh callback or event handler should cause
+ this method to be invoked upon failure, via
+ rd_kafka_oauthbearer_set_token_failure().
+*/
+PHP_METHOD(RdKafka, oauthbearerSetTokenFailure)
+{
+    kafka_object *intern;
+    const char *errstr;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "s", &errstr) == FAILURE) {
```
arnaud-lb: The `s` specifier expects two arguments: a `char*` and a `size_t` (for the length).

rdkafka.c
```diff
@@ -409,6 +409,147 @@ PHP_METHOD(RdKafka, setLogLevel)
...
+    size_t token_value_len;
+    zend_long lifetime_ms;
+    char *principal_name;
+    size_t principal_len;
+    zval *extensions_arg = NULL;
+    
+    char errstr[512];
+    rd_kafka_resp_err_t ret = 0;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "sls|a", &token_value, &token_value_len, &lifetime_ms, &principal_name, &principal_len, &extensions_arg) == FAILURE) {
```
arnaud-lb: Nit pick: use `h` instead of `a` here to get a `HashTable*` directly

arnaud-lb (commented):
Nice!

Do you think it would be feasible to add tests for these methods?

Currently we start a Kafka instance in CI here: https://github.com/arnaud-lb/php-rdkafka/blob/bcd5004f461d1d3a5f879bb21280bdde6f6800c2/.github/workflows/test/start-kafka.sh.

cb-freddysart:
> Nice!
> 
> Do you think it would be feasible to add tests for these methods?
> 
> Currently we start a Kafka instance in CI here: https://github.com/arnaud-lb/php-rdkafka/blob/bcd5004f461d1d3a5f879bb21280bdde6f6800c2/.github/workflows/test/start-kafka.sh.

Thank you for the great review @arnaud-lb . I think I’ve got this all fixed locally but I want to take the time to get valgrind setup to be sure (I’m on a m1 which is not supported yet so I need to break out my Linux machine).

I think I should be able to write some integration tests for this. Hopefully should have something up for review by the end of the week.

stefan-bolsinger-hald:
I'm thrilled that the PR for oauth2 has already been implemented because I urgently need this feature for an application. Is it possible to provide a new release with this feature in the short term? That would be a big help for me.

cb-freddysart:
Hey all, I hope to get this over the line but I've had a shift in priorities after coming back from some time off. Configuring the OAuth side of things on Kafka is new to me and I haven't had the time to figure out exactly how that should work (for example, do we also need to spin up a keycloak container for generating and validating the tokens?).

I'm open to any help if anyone else is more familiar in this area. Otherwise, I'll keep poking at this when I have time.

scorgn:
What left is needed for this MR? Is there anything other than integration tests?

I'm not too familiar with C but I would be happy to help brush up on it and work on implementing integration tests if that is all that is needed for merging.

devenjahnke:
@arnaud-lb What is blocking this PR from being merged? I'm happy to take a stab at rounding it out.

I need this feature for AWS IAM authentication with MSK, and I will be blocked until this merges.

Cheers!

cb-freddysart:
afaik, the only thing blocking this PR is the additional tests that @arnaud-lb asked for. I haven't had the time to figure out how to get kafka set up properly.

scorgn:
Depending on the level of integration testing needed, we may need a new script to set Kafka up with an oauth server in front of it.

@arnaud-lb Can you provide any context to the level of integration testing you would like to see for this MR? Are you looking for a full oauth server implementation or just ensuring that new methods don't break anything with the current kafka implementation?

@devenjahnke Do you have experience with C in order to be able to implement integration testing? I am happy to create a new `start-kafka-with-oauth.sh` script that sets up kafka with an oauth server for the new integration tests. Looking at current integration tests, it looks like I could also implement new ones with the new oauth kafka server in mind as well, but if you have more experience with C you will likely be able to make them easier than I could.

scorgn:
I am in the process of writing tests for it now, I should have a commit soon. There are a couple things that I have fixed so far and a couple things that may still need to be fixed.
 
`$conf->setOauthbearerTokenRefreshCb()` needs to be called before instantiating the Producer or Consumer class with the same `$conf` object in order for it to call the callback successfully. If I instantiate the `Consumer` or `Producer` with a`$conf` object and then directly after set the callback then the callback is not called when the token needs to be refreshed. I'm not sure if this is expected and the Consumers/Producer takes a snapshot of the config when instantiated, though it makes it a bit more complex since you need to access the Consumer/Producer from within the callback function itself.

Can someone chime in on if it is expected that callbacks on the conf object need to be set before a Producer/Consumer is instantiated with the conf object?

The other thing that may not be working as expected is that calling `$client->setOauthbearerTokenFailure()` from within the refresh callback doesn't seem to trigger an error/the error call ack, it needs to be called before the refresh callback. I'm not sure how exactly Kafka is supposed to act when setting the token failure message so I may need to look into this more, but if anyone has any insight on expected behavior when calling '$client->setOauthbearerTokenFailure()' that would be helpful.

scorgn:
@cb-freddysart I’ve submitted a PR to your branch that starts a Kafka container with the oauthbearer security protocol alongside the regular Kafka container, and implements integration testing for the oauthbearer methods. It also includes a few small related fixes in the rdkafka extension. Would you be able to review and merge it into your branch?

https://github.com/cb-freddysart/php-rdkafka/pull/1

Once that's merged, we'll need to resolve merge conflicts. I decided not to merge 6.x into my branch and resolve conflicts beforehand to make the PR into your branch easier to review with fewer changes.

After 6.x is merged into this branch, there’s one additional fix required for a test implemented in 6.x that doesn’t exist yet in this branch. The test is located at `./tests/controller_id.phpt`. It expects the broker ID to be 1001, which is the generated broker ID when none is set. However, my changes manually set the broker ID to 1, so we’ll need to update the test to expect 1 instead of 1001.

I had to manually set the broker IDs in order to have two separate Kafka containers using the same Zookeeper instance, and I’m unable to set the broker ID above 1000. Therefore, we can’t retain the broker ID that the future test expects.

arnaud-lb:
Thanks a lot for the help @scorgn. The PR will be mergeable with these changes.

scorgn:
Happy to help! Thanks for confirming.

@cb-freddysart I also created another PR that does have 6.x merged in, conflicts resolved, and that test fixed. It will have more changes since it also includes commits from 6.x, but I created it just in case it is easier for you to merge in that PR and not have to fix conflicts/test.

https://github.com/cb-freddysart/php-rdkafka/pull/2

Will close whichever one is not merged.

cb-freddysart:
Great work @scorgn! I've approved both PRs. Let's merge the second PR since you have already gone to the trouble of pulling in 6.x?

scorgn:
@cb-freddysart Yup that makes sense! I am not able to merge but you should be able to.

.github/workflows/package.yml
```diff
@@ -20,6 +20,6 @@ jobs:
               run: './php-rdkafka/.github/workflows/package/package.sh'
 
             - name: 'Archive package'
-              uses: 'actions/upload-artifact@v2'
+              uses: 'actions/upload-artifact@v4'
```
cb-freddysart, line 23: See https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

scorgn:
Do we need to add the new oauth test file to package.xml?

cb-freddysart:
@scorgn, it's not totally clear to me as I've only be debugging this through action runs, but it looks like we might have a little flakiness with the kafka setup. I've increased the metadata timeout and enabled debug logs for the broker for the time being. This got the tests to more reliably get past kafka setup, but now we are seeing intermittent failures around topics not existing.

scorgn:
@cb-freddysart Is the flakiness with those tests with v0.11.6 that say unknown broker? If so those are also flaky in 6.x and have been for a little bit. If we want to try to fix those I think a separate PR would be better, but I don't think it needs to be addressed for this PR.

@arnaud-lb Do checks look ok as far as this being mergeable?

eborden:
@arnaud-lb any idea when this will merge and be released? I'm blocked until this gets in.

arnaud-lb:
Thank you @cb-freddysart @scorgn !

arnaud-lb:
Released 6.0.4: https://github.com/arnaud-lb/php-rdkafka/releases/tag/6.0.4

oaattia:
Thank you for adding this PR. I am having difficulty finding an example that demonstrates how to use it for consuming from MSK IAM. I currently have the following code, but unfortunately, it is not working as expected. 

Could someone kindly provide me with some guidance on how to implement it? 

```php
$this->kafkaConfig->setOauthbearerTokenRefreshCb(function ($kafka, $oauthbearer_config = null) {
            try {
                $provider = \Aws\Credentials\CredentialProvider::defaultProvider(
                    [
                        'region'  => env('AWS_DEFAULT_REGION'),
                        'version' => 'latest'
                    ]
                );
                
                $credentials = $provider()->wait();
                
                // using https://github.com/PyaeSoneAungRgn/aws-msk-iam-sasl-signer-php
                $signer = new AwsMskIamSaslSigner(
                    env('AWS_DEFAULT_REGION'),
                    $credentials->getAccessKeyId(),
                    $credentials->getSecretKey(),
                    $credentials->getSecurityToken()
                );
                
                
                $token = $signer->generateToken();

                $kafka->oauthbearerSetToken(
                    $token['token'],
                    $token['expiryTime'],
                    'AWS MSK IAM',
                    []
                );
            } catch (\Exception $e) {
                var_dump('OAuth token generation failed:', $e->getMessage());
                throw new \RuntimeException('Failed to generate OAuth token: ' . $e->getMessage());
            }
        });
```

oaattia:
[2025-01-08 17:13:31] staging.ERROR: Call to undefined method RdKafka\KafkaConsumer::oauthbearerSetToken() {"exception":"[object] (Error(code: 0): Call to undefined method RdKafka\\KafkaConsumer::oauthbearerSetToken() at /var/www/orca/vendor/devmo/kafka-listener/src/Consumer/AwsTopicConsumer.php:123)

oaattia:
Hi, thank you for the PR, @arnaud-lb. 

I have a question about this: Does the OAUTHBEARER only support low-level APIs? I’ve noticed that it hasn't been implemented for high-level APIs as well

PR #568, opened 2024-10-30
Fix segmentation fault in setOauthbearerTokenRefreshCb when sasl.oauthbearer.config is unset (#567)

### LINKED ISSUE #567: Segmentation fault when using setOauthbearerTokenRefreshCb without setting sasl.oauthbearer.config
### Description
Using `setOauthbearerTokenRefreshCb` without setting `sasl.oauthbearer.config` on `\RdKafka\Conf` results in a segmentation fault since the extension is trying to call `ZVAL_STRING(&args[1], oauthbearer_config);` when `oauthbearer_config` is null.

The line that the seg fault occurs at is: [conf.c#363](https://github.com/arnaud-lb/php-rdkafka/blob/7.x/conf.c#L363).

To fix this we should not call `ZVAL_STRING(&args[1], oauthbearer_config);` when `oauthbearer_config` is null. 

We already call `ZVAL_NULL(&args[1]);` on line 360, so we can just wrap line 363 in an if statement without an else.

I created a new test file [bug567.phpt](https://github.com/scorgn/php-rdkafka/pull/2/files#diff-87740d2e31f3edb12f00cd814a245b9bbe0ace0d6a354acae81ae7bd1591fb66) and created an MR into my own branch without a fix in order to have actions running and show them failing. Here is the [example MR showing actions failing](https://github.com/scorgn/php-rdkafka/pull/2)

I implemented a fix in this MR which show those tests passing:
https://github.com/arnaud-lb/php-rdkafka/pull/568

The following code:

```php
<?php

$conf = new RdKafka\Conf();
if (RD_KAFKA_VERSION >= 0x090000 && false !== getenv('TEST_KAFKA_BROKER_VERSION')) {
    $conf->set('broker.version.fallback', getenv('TEST_KAFKA_BROKER_VERSION'));
}

$conf->set('metadata.broker.list', 'foobar');
$conf->set('security.protocol', 'SASL_PLAINTEXT');
$conf->set('sasl.mechanisms', 'OAUTHBEARER');
$conf->setOauthbearerTokenRefreshCb(function ($producer) {
    echo "oauthbearer token refresh callback successfully called\n";
});
$producer = new \RdKafka\Producer($conf);
$producer->poll(0);
echo "producer polled successfully\n";

```

Resulted in a segmentation error

But I expected this output instead:
```
oauthbearer token refresh callback successfully called
producer polled successfully
```

### php-rdkafka Version

v6.0.4

### librdkafka Version

_No response_

### PHP Version

_No response_

### Operating System

_No response_

### Kafka Version

_No response_

scorgn:
Only call `ZVAL_STRING` for `&args[1]` with `oauthbearer_config` if `oauthbearer_config` is set in `kafka_conf_set_oauthbearer_token_refresh_cb`

Fixes #567

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -365,7 +365,10 @@ static void kafka_conf_set_oauthbearer_token_refresh_cb(rd_kafka_t *rk, const ch
     ZVAL_NULL(&args[1]);
 
     ZVAL_ZVAL(&args[0], &cbs->zrk, 1, 0);
-    ZVAL_STRING(&args[1], oauthbearer_config);
+
+    if (oauthbearer_config) {
+        ZVAL_STRING(&args[1], oauthbearer_config);
+    }
 
     rdkafka_call_function(&cbs->oauthbearer_token_refresh->fci, &cbs->oauthbearer_token_refresh->fcc, NULL, 2, args);
 
--- a/package.xml
+++ b/package.xml
@@ -118,6 +118,7 @@
     <file role="test" name="bug465.phpt"/>
     <file role="test" name="bug508.phpt"/>
     <file role="test" name="bug521.phpt"/>
+    <file role="test" name="bug567.phpt"/>
     <file role="test" name="bug74.phpt"/>
     <file role="test" name="bug88.phpt"/>
     <file role="test" name="bugConfSetArgument.phpt"/>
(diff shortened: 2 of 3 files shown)


arnaud-lb:
Thank you!

PR #576, opened 2025-02-24
Add rdkafka to the include path

arnaud-lb:
Should fix https://github.com/arnaud-lb/php-rdkafka/issues/575.


ralphschindler:
Closing as per comment on original issue: https://github.com/php-rdkafka/php-rdkafka/issues/575#issuecomment-2681891009

PR #582, opened 2025-05-19
Fix missing call to zend_restore_error_handling()

ndossche:
Found with my experimental static analyzer

### DIFF
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -145,6 +145,7 @@ PHP_METHOD(RdKafka__KafkaConsumer, __construct)
     }
 
     if (!has_group_id(conf)) {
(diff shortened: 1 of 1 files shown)


arnaud-lb:
Thanks you!

Merged in 6.x and up: https://github.com/arnaud-lb/php-rdkafka/commit/4e0d407adcc0dd4ea71c4ec22caba5ba644d1035

ndossche:
FWIW, the reason I targeted 5.x was because of this line in the contributing docs:

> Pull requests should be made against the 5.x branch, which supports both PHP 7 and PHP 8.

So perhaps that needs updating? :thinking:

arnaud-lb:
Ah thank you! I will update this

Issue #454, opened 2021-04-22
Support to alter offsets


KlimTodrik:
* PHP version: any
 * librdkafka version: any
 * php-rdkafka version: 1.6.1
 * kafka version: 0.10+

Can you create support for [seek](https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#seek(org.apache.kafka.common.TopicPartition,%20long)) method? 

It allows to shift offset on consumer. 

As you can see this method alerady [implemented](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.c#L2746) in rdkafka

nick-zh:
You can read from specific offsets using `KafkaConsumer:assign`.
Also as mentioned in Gitter, if you really want to change the offset, use the scripts provided with Kafka, in this case `kafka-consumer-groups.sh`.

KlimTodrik:
Yes I know about CLI, but I need to update offsets via rdkafka

nick-zh:
As i said in Gitter, from what i understand, neither `assign` nor `seek` changes the commited offset on Kafka, but simply allows the consumer to read from a previous offset. If this is all you need to do, then `assign` should do the job.

KlimTodrik:
No, seek allows to shift offset for selected group. Like CLI `--reset-offsets --to-offset`

nick-zh:
I am fairly certain it does not, the [doc block](https://github.com/edenhill/librdkafka/blob/8bacbc0b4c357193288c81277bfcc815633126ea/src/rdkafka.h#L3499) in librdkafka suggest that it will simply refetch messges from an offset, but it does not mention anything about changing the offset. [This](https://github.com/edenhill/librdkafka/issues/2965#issuecomment-654198538) comment from Magnus suggests the same. I am open though if you can point me to a source which mentions that the commited offset is actually changed.

nick-zh:
What you are looking for can only be achieved with the AdminApi from what i know and we do not support that yet (see #215)

nick-zh:
I am closing this, since what you want is a possibility to alter offsets and as mentioned, we'll solve this in #215 :v:

Issue #385, opened 2020-09-04
Not able to delete topic using  php-rdkafka library 


rsarkar29:
Hi I am not getting any function to delete topics, is there any function to delete topics from Kafka using this library?

nick-zh:
This is not supported yet, check #215 for updates. It is planned for https://github.com/arnaud-lb/php-rdkafka/milestone/2

Issue #335, opened 2020-01-07
define amount of replications and partitions


chrisfield490:
* PHP version: 7.3.11
 * librdkafka version:
 * php-rdkafka version:
 * kafka version: 2.4.0

Steveb-p:
Php-rdkafka does not yet have ability to use Admin API of librdkafka. See #215.

chrisfield490:
Noted Thank you

Issue #243, opened 2019-07-24
when i call func newTopic,how to set partition number?and how many default partitions


ayamzh:
$rk = new \RdKafka\Producer();
        $rk->setLogLevel(LOG_DEBUG);
        $rk->addBrokers("127.0.0.1:9092");

        $topic = $rk->newTopic("test");

nick-zh:
@ayamzh when you create a topic on the broker, you define how many partitions you want. The default setting on the broker, which is `num.partitions` is set to 1.
With either `produce` and `producev` you can either give a partition you want, or set `RD_KAFKA_PARTITION_UA` (unassigned) which will take the default partitioner. As far as i remember the default is a random partitioner, but you could set others with `RdKafka\TopicConf::setPartitioner`.
I hope this helps :)

ayamzh:
can you give me an example  code,hot to use php-rdkafka  set the partition number of the new topic

nick-zh:
@ayamzh i am not sure if i understand correct. You cannot create the topic / partition with PHP since the Admin API is not exposed (https://github.com/arnaud-lb/php-rdkafka/issues/215). `newTopic` does not create a new topic, it creates an instance of a topic that you need for the consumer.

nick-zh:
@ayamzh let me know if you need further help or if we can resolve this

ayamzh:
thanks very much,  i just see #215 , the struct has  "num_partitions" 
/**
 * @name CreateTopics
 * @{
 */

/**
 * @brief NewTopic type, used with CreateTopics.
 */
struct rd_kafka_NewTopic_s {
        /* Required */
        char *topic;            /**< Topic to be created */
        int num_partitions;     /**< Number of partitions to create */
        int replication_factor; /**< Replication factor */

        /* Optional */
        rd_list_t replicas;     /**< Type (rd_list_t (int32_t)):
                                 *   Array of replica lists indexed by
                                 *   partition, size num_partitions. */
        rd_list_t config;       /**< Type (rd_kafka_ConfigEntry_t *):
                                 *   List of configuration entries */
};

Steveb-p:
@ayamzh You're referencing librdkafka API that Nick specifically said earlier is not implemented yet in phprdkafka (#215)

ayamzh:
the kafka  config/server.properties's num.partitions , is the partitions default value when i use php client to call newTopic,  but

ayamzh:
[...]

PR #592, opened 2026-02-25
7.x - Implement AdminClient with topic management operations

aknEvrnky:
## Summary

Adds a new `RdKafka\Admin\AdminClient` class that exposes librdkafka's admin API to PHP, covering the 5 most-requested topic management operations:

- **CreateTopics** — create topics with partition count, replication factor, and optional config/replica assignment
- **DeleteTopics** — delete topics by name
- **CreatePartitions** — increase partition count for existing topics
- **DescribeTopics** — get topic metadata including partition info, leader/ISR/replica nodes
- **DeleteRecords** — delete records before a given offset per partition

### New classes

| Class | Purpose |
|---|---|
| `RdKafka\Admin\AdminClient` | Main entry point, wraps a producer-type `rd_kafka_t` handle |
| `RdKafka\Admin\AdminOptions` | Per-operation options (timeouts, validate-only, broker ID) |
| `RdKafka\Admin\NewTopic` | Input for `createTopics()` |
| `RdKafka\Admin\DeleteTopic` | Input for `deleteTopics()` |
| `RdKafka\Admin\NewPartitions` | Input for `createPartitions()` |
| `RdKafka\Admin\TopicResult` | Result for create/delete operations |
| `RdKafka\Admin\TopicDescription` | Result for `describeTopics()` |
| `RdKafka\Admin\TopicPartitionInfo` | Partition detail within a TopicDescription |
| `RdKafka\Admin\Node` | Broker node info (id, host, port, rack) |

### Design decisions

- **Synchronous PHP API** wrapping librdkafka's async queue-based pattern internally — the extension creates a temporary queue, calls the operation, polls for the result, and returns it directly.
- **`AdminOptions` via factory method** (`$admin->newAdminOptions(RD_KAFKA_ADMIN_OP_*)`) because the underlying C API requires the `rd_kafka_t` handle.
- **`describeTopics()` takes `string[]`** instead of exposing a `TopicCollection` wrapper class — keeps the API simple while the collection is built internally.
- **`deleteRecords()` takes `TopicPartition[]`** — reuses the existing `RdKafka\TopicPartition` class and internal helpers.
- Follows all naming/design conventions from CONTRIBUTING.md.

### Tests

- 1 unit test (`admin_client_001.phpt`) — object creation and configuration without a broker
- 4 integration tests against a running broker:
  - `admin_client_create_delete_topics.phpt`
  - `admin_client_describe_topics.phpt`
  - `admin_client_create_partitions.phpt`
  - `admin_client_delete_records.phpt`

### Related issues

Addresses the topic management portion of:
- #215 — AdminClient API support (main tracking issue)
- #243 — Create topic programmatically
- #335 — Dynamically create topics
- #385 — Delete topic

Consumer group operations (#454) and remaining admin APIs can follow as separate PRs.


emir:
👍

ralphschindler:
A couple of changes (small, timeout, etc), a rebase to get the tests working in CI.

ralphschindler:
Thanks for this... the struct wrappers (`NewTopic`, `DeleteTopic`, `AdminOptions`, etc.) are solid and map cleanly to the librdkafka types.

Before we merge, I want to raise a design question and get some other eyes on it. The extension's stated goal is to be a thin wrapper around librdkafka, following its API as closely as possible. This PR makes two choices that sit a bit outside that:

1. **The synchronous abstraction**: librdkafka's admin API is async and queue-based. You submit an operation and poll a queue for the result. The current implementation hides that entirely, blocking internally until the event arrives. The rest of the extension exposes timeouts and polling directly to the caller (e.g. `consume()`, `poll()`). A more faithful approach might be `createTopics(array $topics, Queue $queue, ?AdminOptions $options)` and let the caller drive the event loop.
2. **`AdminClient` as a standalone class**: `librdkafka` doesn't have a separate admin handle; admin functions operate on any `rd_kafka_t` producer handle. The `AdminClient` class is an invention borrowed from the Java/Python higher-level clients, not from the C API itself.

Neither of these is necessarily wrong, but they're meaningful departures from the thin-wrapper philosophy and worth a deliberate decision rather than just landing quietly.

@arnaud-lb @nick-zh I would love your take on the right direction here before we move forward.

arnaud-lb:
The synchronous abstraction simplifies the API, but would be an issue in async environments. It would be better to stick to the thin-wrapper philosophy at least for this reason. It is possible to create a simpler synchronous API in a PHP package on top of the extension.

Regarding AdminClient, I would also prefer that methods are exposed on RdKafka\Kafka directly, unless the separation is required for technical reasons.

aknEvrnky:
Thanks for the thoughtful review, @ralphschindler and @arnaud-lb.

When I initially designed this, I studied the Java and Python clients, where `AdminClient` is a standalone class with a synchronous-style API. That influenced both the class separation and the blocking approach. However, those SDKs are higher-level abstractions, whereas php-rdkafka's philosophy is to stay as close to librdkafka's C API as possible. So I agree with both points.

Here's what I'm thinking, and I'd appreciate your confirmation before moving forward:

1 - **Move admin methods onto `RdKafka\Kafka` directly** and remove the separate `AdminClient` class, since librdkafka operates on any `rd_kafka_t` handle. The `AdminOptions` and result/input value objects (NewTopic, DeleteTopic, NewPartitions, TopicResult, etc.) would still live under the `RdKafka\Admin` namespace as plain data containers.

2 - **Expose the queue-based async pattern** instead of blocking internally, methods would take a `Queue` parameter and let the caller drive polling, consistent with the rest of the extension. A synchronous convenience layer can always be built on top in userland PHP.

To make the shape of this concrete, here's how the API would look before and after this change:

**Before (current PR):**

```php
$admin = new RdKafka\Admin\AdminClient($conf);
$options = $admin->newAdminOptions(RD_KAFKA_ADMIN_OP_CREATETOPICS);
$newTopic = new RdKafka\Admin\NewTopic('my-topic', 3, 1);

// blocks internally until result arrives
$results = $admin->createTopics([$newTopic], $options);
```

**After (proposed direction):**

```php
$producer = new RdKafka\Producer($conf); // no separate AdminClient
$queue = $producer->newQueue();
$options = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_CREATETOPICS);
$newTopic = new RdKafka\Admin\NewTopic('my-topic', 3, 1);
// submits the operation; does not block
$producer->createTopics([$newTopic], $queue, $options);
// caller drives the event loop
$event = $queue->poll($timeoutMs);
$results = $event->getCreateTopicsResult();
```

One thing this surfaces: today `RdKafka\Queue` only exposes `consume() `which returns a `Message`. To support admin events properly, we'd need to add:

- `Queue::poll(int $timeout_ms): ?Event` — mirroring `rd_kafka_queue_poll()`
- A new `RdKafka\Event` class wrapping `rd_kafka_event_t`, with type-specific result accessors

There are a couple of ways we could shape Event. One is a single class with `getType()` and `get*Result()` methods, which stays closest to the C API. The other is type-specific subclasses like `CreateTopicsEvent`, `DeleteTopicsEvent`, and so on, which is more idiomatic in PHP. I'd lean toward the first for consistency with librdkafka, but open to either.

If this direction works for you, I can pick up the issue and implement it.

ralphschindler:
I think this direction works and is consistent with the project philosophy and satisfies @arnaud-lb's preferred direction insofar as I can tell. Let's go in this direction (polling pushed to userland, admin API methods on Kafka class, and keeping `RdKafka\Admin` namespace for dto/value objects). We can document the way to use this, and what a userland implementation of the async poll loop. Thanks!

aknEvrnky:
## Admin API Refactor Summary

This update aligns the Admin API with php-rdkafka's thin-wrapper philosophy and the maintainer feedback on PR #592.

### What changed and why

1. **Removed `RdKafka\Admin\AdminClient`**
   - **What changed:** Admin entry points are no longer exposed through a separate `AdminClient` class.
   - **Why:** librdkafka does not have a separate admin handle. Admin operations run on a regular `rd_kafka_t` handle.

2. **Moved admin methods onto `RdKafka` (usable from `Producer`/`Consumer`)**
   - **What changed:** `newQueue()`, `newAdminOptions()`, and admin methods (`createTopics`, `deleteTopics`, `createPartitions`, `describeTopics`, `deleteRecords`) are now on `RdKafka`.
   - **Why:** This matches librdkafka's handle-based API shape and removes an unnecessary abstraction layer.

3. **Switched from internal blocking to queue/event async flow**
   - **What changed:** Admin methods now submit requests and return immediately. Caller provides a `Queue`, then polls it.
   - **Why:** This is consistent with the extension's existing poll-driven model and avoids hidden blocking, especially important for async runtimes.

4. **Added `RdKafka\Queue::poll(int $timeout_ms): ?RdKafka\Event`**
   - **What changed:** `Queue` can now poll for generic librdkafka events (not only message consume flow).
   - **Why:** Admin results are delivered as queue events in librdkafka.

5. **Added new `RdKafka\Event` wrapper**
   - **What changed:** `Event` exposes:
     - `getType()`
     - `getName()`
     - `getError()`
     - `getErrorString()`
     - per-operation result accessors
   - **Why:** This maps directly to librdkafka's `rd_kafka_event_t` model.

6. **Kept `RdKafka\Admin` namespace for DTO/value objects**
   - **What changed:** `AdminOptions`, `NewTopic`, `DeleteTopic`, `NewPartitions`, `TopicResult` (and describe-related DTOs) remain under `RdKafka\Admin`.
   - **Why:** These are still valid value/object mappings for admin inputs and outputs.

7. **Updated tests, docs, and build/packaging**
   - **What changed:** PHPT tests rewritten for queue/event API, README updated, and build manifests updated (`config.m4`, `config.w32`, `package.xml`).
   - **Why:** Ensure CI/package correctness and make new API discoverable/documented.

8. **Version compatibility preserved**
   - **What changed:** `describeTopics` and related APIs remain gated by `HAS_RD_KAFKA_DESCRIBE_TOPICS`.
   - **Why:** Keep compatibility with older librdkafka versions where DescribeTopics is unavailable.

---

## Usage Guide (Mini README)

### 1) Setup

```php
<?php

$conf = new RdKafka\Conf();
$conf->set('bootstrap.servers', '127.0.0.1:9092');

$queue = $producer->newQueue();
```

### 2) Create topics

```php
$opts = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_CREATETOPICS);
$opts->setRequestTimeout(5000);

$producer->createTopics(
    [new RdKafka\Admin\NewTopic('my-topic', 3, 1)],
    $queue,
    $opts
);

$event = $queue->poll(10000);
if ($event === null) {
    throw new RuntimeException('Timed out waiting for CreateTopics result');
}
if ($event->getError() !== RD_KAFKA_RESP_ERR_NO_ERROR) {
    throw new RuntimeException($event->getErrorString() ?? 'CreateTopics event error');
}

foreach ($event->getCreateTopicsResult() as $result) {
    printf("%s => error=%d, errorString=%s\n",
        $result->getName(),
        $result->getError(),
        var_export($result->getErrorString(), true)
    );
}
```

### 3) Delete topics

```php
$opts = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_DELETETOPICS);
$opts->setRequestTimeout(5000);

$producer->deleteTopics(
    [new RdKafka\Admin\DeleteTopic('my-topic')],
    $queue,
    $opts
);

$event = $queue->poll(10000);
if ($event === null) {
    throw new RuntimeException('Timed out waiting for DeleteTopics result');
}

$results = $event->getDeleteTopicsResult();
```

### 4) Create partitions

```php
$opts = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_CREATEPARTITIONS);
$opts->setRequestTimeout(5000);

$producer->createPartitions(
    [new RdKafka\Admin\NewPartitions('my-topic', 6)],
    $queue,
    $opts
);

$event = $queue->poll(10000);
if ($event === null) {
    throw new RuntimeException('Timed out waiting for CreatePartitions result');
}

$results = $event->getCreatePartitionsResult();
```

### 5) Describe topics (when supported by librdkafka)

```php
if (method_exists(RdKafka::class, 'describeTopics')) {
    $opts = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_DESCRIBETOPICS);
    $opts->setRequestTimeout(5000);

    $producer->describeTopics(['my-topic'], $queue, $opts);

    $event = $queue->poll(10000);
    if ($event === null) {
        throw new RuntimeException('Timed out waiting for DescribeTopics result');
    }

    $descriptions = $event->getDescribeTopicsResult();
    foreach ($descriptions as $topicDesc) {
        echo $topicDesc->name . PHP_EOL;
    }
}
```

### 6) Delete records

```php
$opts = $producer->newAdminOptions(RD_KAFKA_ADMIN_OP_DELETERECORDS);
$opts->setRequestTimeout(10000);

$producer->deleteRecords(
    [new RdKafka\TopicPartition('my-topic', 0, 1234)],
    $queue,
    $opts
);

$event = $queue->poll(15000);
if ($event === null) {
    throw new RuntimeException('Timed out waiting for DeleteRecords result');
}

$offsetResults = $event->getDeleteRecordsResult(); // TopicPartition[]
```

### 7) Optional userland sync helper

```php
function adminSubmitAndWait(callable $submit, RdKafka\Queue $queue, int $timeoutMs = 10000): RdKafka\Event
{
    $submit();

    $event = $queue->poll($timeoutMs);
    if ($event === null) {
        throw new RuntimeException("Admin operation timed out after {$timeoutMs}ms");
    }

    if ($event->getError() !== RD_KAFKA_RESP_ERR_NO_ERROR) {
        throw new RuntimeException($event->getErrorString() ?? 'admin event error');
    }

    return $event;
}
```

ralphschindler:
Finally got a proper block of time to dig into the current state of this — the redesign looks great. Methods on `RdKafka` directly, the queue/event pattern, `RdKafka\Admin` kept to DTOs only — this is exactly the shape the thin-wrapper philosophy calls for, and it's a clean, faithful mapping of librdkafka's actual admin API. Nice work carrying that through.

I rebased locally onto current `7.x` to review a clean diff (this branch has drifted a few months behind — conflicts were trivial, all just parallel additions to the same config.m4/arginfo lines, so that part isn't a concern). A few things turned up that I'd want fixed before merging:

- **`RdKafka\Event` needs to hold a reference back to its owning `RdKafka` handle**, the same way `Queue` and `AdminOptions` already do. librdkafka's own docs list `rd_kafka_event_t` among the objects that *must* be destroyed before the handle is — without that reference, something like `(new Producer($conf))->newQueue()->poll(10000)` can destroy the producer while the returned `Event` is still alive, which per their docs risks `rd_kafka_destroy()` hanging. `Queue`'s existing `zrk` field is the pattern to copy.
- **`admin_client.c`/`admin_client.h`/`event.c` use `XtOffsetOf`**, which was removed from php-src and will break PHP 8.6 builds — same issue #612 just fixed everywhere else, just not here since these files didn't exist yet at the time. Swapping to `offsetof` (or better, reusing the existing `php_kafka_from_obj` macro) fixes it.
- **`NewTopic`/`NewPartitions` constructors cast user ints straight to `int`/`size_t`** with no negative check, so e.g. `new NewPartitions('t', -1)` sends `SIZE_MAX` into librdkafka instead of throwing cleanly.
- A handful of `object_init_ex()` calls don't check their return value (inconsistent with the rest of the codebase, and with this PR's own `kafka_event_new`) — `newAdminOptions` also leaks the already-created `AdminOptions` on that failure path.
- Smaller stuff: `setReplicaAssignment` skips the empty-array guard the rest of the admin methods have, and silently coerces non-integer broker IDs; the cross-handle rejection check isn't covered by a test; some duplicated loops (`createTopics`/`deleteTopics`/`createPartitions` could follow `deleteRecords`'s lead and reuse a shared helper).

Given how much of this is yours already and how well the redesign turned out, I'd rather this go back to you than have me take it over — happy to have the attribution land where it's earned. Want to take a pass at these yourself, or would you rather I pick them up? Either way works for me, just let me know. The `Event` lifetime one is the only genuinely non-obvious one (needs librdkafka's termination-order contract, not just PHP conventions) — happy to pair on that specific piece if useful.

aknEvrnky:
Thank you for your feedback. I'd like to finish what I started and finalize it. I already rebased the `7.x` branch and fixed conflict. Currently working on your feedbacks. 

I'll let you know if I need any support.

aknEvrnky:
I went through the list on this branch.

- `Event` now holds a `zrk` ref to the owning handle so `rd_kafka_destroy()` can’t run first.
- Remaining `XtOffsetOf` in the Admin files swapped to `offsetof` / `php_kafka_from_obj`.
- `NewPartitions` rejects negatives; `setReplicaAssignment` rejects empty arrays and non-int broker IDs. `NewTopic(-1, -1)` is still allowed.
- `object_init_ex` is checked; `newAdminOptions` no longer leaks on that path.
`createTopics` / `deleteTopics` / `createPartitions` share one array collector.
- Describe-topics conversion now dtors in-progress zvals on exception.

Added `admin_003_cross_handle.phpt` plus DTO / Event-lifetime coverage. Local `make test` is green (OAuth skipped). The 6 red CI jobs are memcheck apt-get 404s for libc6-dbg, not the suite.

I think this PR is ready to be merged.

Issue #382, opened 2020-08-26
Support for transactional consume-transform-produce loop


nick-zh:
For the use case where you have the following:
- consumer reading messages
- processing those messages
- producing a result of those processed messages
- all the above being part of a transaction

We need to be able for the consumer offsets to be part of the transaction, this can be achieved by using `rd_kafka_send_offsets_to_transaction` for which we should add support for.

Is related to #359

donatascn:
It is odd that this feature is not being upvoted :)

nizacegodk:
We would also benefit from having this feature in our projects.

ralphschindler:
Added with #597

PR #597, opened 2026-04-23
transactional consume-transform-produce loop

ralphschindler:
Would close feature request in #382

- new `RdKafka\ConsumerGroupMetadata`
- `KafkaConsumer::getConsumerGroupMetadata()`
- `Producer::sendOffsetsToTransaction(array $offsets, ConsumerGroupMetadata $metadata, int $timeout_ms)`

The getter methods (`getGroupId(), getMemberId(), getGroupInstanceId(), getGenerationId()`) are always present on the class but require `librdkafka` >= 2.0.0 to function. When called against an older version they throw `RdKafka\Exception` with a message indicating the version requirement. The
  constructor and `getConsumerGroupMetadata()` work on all supported versions (librdkafka 1.5.3+).


ralphschindler:
Hi @nick-zh do you happen to have the capacity to test this out? Thanks!

PR #599, opened 2026-04-24
Add Conf::get() and TopicConf::get()

### LINKED ISSUE #525: Getting librdkafka builtin features
### Problem

**builtin.features:** gzip, snappy, ssl, sasl, regex, lz4, sasl_gssapi, sasl_plain, sasl_scram, plugins, zstd, sasl_oauthbearer, http, oidc.
https://github.com/edenhill/librdkafka/blob/b47da0e4eec900d9cb091d2f524e957505dbb513/CONFIGURATION.md
It is not possible to detect the available features of librdkafka using custom PHP code. I can't dynamically manage librdkafka options depending on the environment.

### Solution
- librdkafka build information in phpinfo.
- There is a constant with librdkafka build information.
- There is a check function (???).

**Hint:** `builtin_features`/`BUILT_WITH` https://github.com/edenhill/librdkafka/blob/b47da0e4eec900d9cb091d2f524e957505dbb513/src/rdkafka.c#L2544

### LINKED ISSUE #527: Add RdKafka\Conf::get() method
### Description

Currently, the only way to read configuration values from RdKafka\Conf or RdKafka\TopicConf is to call the dump() method. It would be convenient if we also added a get() method to fetch a single setting.

See https://github.com/edenhill/librdkafka/blob/a47c74fdc828f332e5136639f166d33c14953e75/src/rdkafka.h#L2446-L2481

ralphschindler:
Adds a `get(string $name)`: string method to both `RdKafka\Conf` and `RdKafka\TopicConf`, wrapping `rd_kafka_conf_get()` and `rd_kafka_topic_conf_get()` respectively.

Currently the only way to read back a config value is dump(), which returns everything. This gives you a single property without the overhead of dumping the full config.

Throws `RdKafka\Exception` for unknown property names, consistent with how set() handles unknown properties.

Also note: like set(), `rd_kafka_conf_get()` always returns string values regardless of the underlying type.

Fixes #527 
Fixes #525

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -433,6 +433,58 @@ PHP_METHOD(RdKafka_Conf, dump)
 }
 /* }}} */
 
+/* {{{ proto string RdKafka\Conf::get(string $name)
+   Retrieve a configuration property value. */
+PHP_METHOD(RdKafka_Conf, get)
+{
+    char *name;
+    size_t name_len;
+    kafka_conf_object *intern;
+    char *dest;
+    size_t dest_size;
+    rd_kafka_conf_res_t ret;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "s", &name, &name_len) == FAILURE) {
+        return;
+    }
+
+    intern = get_kafka_conf_object(getThis());
+    if (!intern) {
+        return;
+    }
+
+    switch (intern->type) {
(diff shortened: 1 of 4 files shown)


ralphschindler:
Closing in favor of #602

PR #600, opened 2026-04-24
KafkaConsumer::consume() returns null on poll timeout (breaking change)

### LINKED ISSUE #343: Improve RD_KAFKA_RESP_ERR__TIMED_OUT to be able to differentiate cases

ralphschindler:
Previously, `KafkaConsumer::consume()` returned a Message object with `err` set to `RD_KAFKA_RESP_ERR__TIMED_OUT` (-185) when no message arrived within the timeout window. This was a php-rdkafka-ism... librdkafka itself returns `NULL` in that case.

This made it impossible to distinguish "nothing arrived during the poll window" from an actual `RD_KAFKA_RESP_ERR__TIMED_OUT` error originating from librdkafka. The return type is now `?Message`, returning `null` when the poll window expires with no message.

Note: A test for a real `RD_KAFKA_RESP_ERR__TIMED_OUT` coming back inside a Message isn't included here... reliably triggering a broker-level timeout in a test environment isn't practical without introducing flakiness.

Before:
```php
$msg = $consumer->consume(1000);
if ($msg->err === RD_KAFKA_RESP_ERR__TIMED_OUT) {
    // no message — but was this a real timeout error or just an empty poll?
}
```

After:
```php
$msg = $consumer->consume(1000);
if ($msg === null) {
    // no message in the poll window
}
```

`RD_KAFKA_RESP_ERR__TIMED_OUT` on a returned `Message` now means an actual timeout error from librdkafka, not a poll window expiry.

Existing tests updated: Any test previously checking `$msg->err === RD_KAFKA_RESP_ERR__TIMED_OUT` (or -185) to detect an empty poll has been updated to check `$msg === null` instead.

Closes #343.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,12 @@ package.xml
 rdkafka-*.tgz
 run-tests.php
 gen_stub.php
+tests/*.diff
+tests/*.exp
+tests/*.log
+tests/*.out
+tests/*.php
+tests/*.sh
 tests/*/*.diff
 tests/*/*.exp
 tests/*/*.log
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -16,6 +16,28 @@
 
 ## User-impacting changes
 
+### `KafkaConsumer::consume()` returns `null` on timeout
+
+Previously, `KafkaConsumer::consume()` returned a `Message` object with `err` set to `RD_KAFKA_RESP_ERR__TIMED_OUT` when no message arrived within the timeout window. It now returns `null` in that case, matching librdkafka's own behavior.
+
+**Before:**
+```php
+$msg = $consumer->consume(1000);
+if ($msg->err === RD_KAFKA_RESP_ERR__TIMED_OUT) {
+    // no message
+}
+```
+
+**After:**
+```php
+$msg = $consumer->consume(1000);
+if ($msg === null) {
+    // no message
+}
+```
+
+`RD_KAFKA_RESP_ERR__TIMED_OUT` on a returned `Message` now means an actual timeout error originating from librdkafka, not a poll window expiry.
+
 ### PHP 8.1 now required
 
 php-rdkafka 7.x requires PHP 8.1 or later. PHP 7.x is no longer supported.
--- a/kafka_consumer.c
+++ b/kafka_consumer.c
@@ -392,13 +392,13 @@ PHP_METHOD(RdKafka_KafkaConsumer, unsubscribe)
 }
 /* }}} */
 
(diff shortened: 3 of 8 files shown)


ralphschindler:
@nick-zh you commented on the original issue about potentially solving this, since we are up to 7 which could include breaking changes, I wonder what your thoughts on this. Thanks!

nick-zh:
The change would help differentiate the signals now, but will overlap with EOF, unless `enable.partition.eof` is set to true. I guess if it is mentioned in the release, it could be ok

ralphschindler:
Good call on noting the EOF overlap... it's worth documenting clearly for anyone landing here.

This change actually brings `consume()` closer to librdkafka's own behavior. `rd_kafka_consumer_poll()` returns `NULL` when there's nothing to deliver.  The previous extension behavior was synthesizing a `Message` object with `err=-185` on top of that, which was an abstraction beyond what librdkafka itself provides. Returning `null` is the more faithful, thinner mapping.

On the EOF overlap: when `enable.partition.eof=false` (the default), `rd_kafka_consumer_poll()` returns `NULL` both on a genuine poll timeout and when a partition is fully caught up to its end. Both result in `consume()` returning `null`. The previous behavior had the same underlying ambiguity as it returned `err=-185` in both cases, since librdkafka doesn't synthesize a separate EOF signal in that mode.

For consumers that need to distinguish the two cases, setting `enable.partition.eof=true` gives you clean, distinct signals:

  ```php
  consume(1000) → null                 // genuine timeout, no data
  consume(1000) → Message{ err: -191 } // partition caught up to EOF
  consume(1000) → Message{ err: 0 }    // real message
```

For the majority of streaming consumers that just want to loop until a message arrives, null is straightforward — it means "nothing right now, try again." The EOF/timeout overlap and the options for handling it are called out in the UPGRADE notes.

PR #602, opened 2026-04-26
Include topic-level properties in Conf::dump()

### LINKED ISSUE #306: Improve RdKafka\Conf::dump
When dumping `RdKafka\Conf` you only see the global settings.
Every `rd_kafka_conf_t`  can also have a `rd_kafka_topic_conf_t`,
so it might be confusing to not find certain settings in the dump,
that were set through `RdKafka\Conf::set` like `auto.offset.reset`.
The behaviour should probably be improved, so that if a `rd_kafka_topic_conf_t` exists, it should be dumped as well.

ralphschindler:
After feedback from @arnaud-lb, this takes a different direction than originally proposed.

Instead of silently merging topic-level properties into `Conf::dump()`, this PR:

- Adds `Conf::get(string $name): string` and `TopicConf::get(string $name): string` — fetch a single config value directly, throwing `RdKafka\Exception` for unknown properties
- Adds `Conf::getDefaultTopicConf(): ?TopicConf` — exposes the embedded default topic conf so topic-level properties can be read explicitly via `->dump()` or `->get()`

`dump()` behavior is unchanged. The pattern for reading topic-level properties is now:

```php
$conf->set('auto.offset.reset', 'earliest');

$conf->get('group.id');                             // global property
$conf->getDefaultTopicConf()->get('auto.offset.reset'); // topic-level property
```

Closes #306, #527, #525.

~~`Conf::dump()` was only dumping the global `rd_kafka_conf_t` settings. Properties like `auto.offset.reset` are routed by librdkafka into the embedded `default_topic_conf`, so they were silently missing from the output... which made dump() unreliable for debugging.~~

~~The fix calls `rd_kafka_conf_get_default_topic_conf()` after the global dump and, if a default topic conf exists, merges its entries into the same array. `TopicConf::dump()` is unchanged.~~

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -410,26 +410,38 @@ PHP_METHOD(RdKafka_Conf, dump)
         return;
     }
 
+    array_init(return_value);
+
     switch (intern->type) {
-        case KAFKA_CONF:
+        case KAFKA_CONF: {
+            rd_kafka_topic_conf_t *topic_conf;
+
             dump = rd_kafka_conf_dump(intern->u.conf, &cntp);
+            for (i = 0; i < cntp; i+=2) {
+                add_assoc_string(return_value, (char*)dump[i], (char*)dump[i+1]);
+            }
+            rd_kafka_conf_dump_free(dump, cntp);
+
+            topic_conf = rd_kafka_conf_get_default_topic_conf(intern->u.conf);
+            if (topic_conf) {
+                dump = rd_kafka_topic_conf_dump(topic_conf, &cntp);
+                for (i = 0; i < cntp; i+=2) {
+                    add_assoc_string(return_value, (char*)dump[i], (char*)dump[i+1]);
+                }
+                rd_kafka_conf_dump_free(dump, cntp);
+            }
             break;
+        }
         case KAFKA_TOPIC_CONF:
             dump = rd_kafka_topic_conf_dump(intern->u.topic_conf, &cntp);
+            for (i = 0; i < cntp; i+=2) {
+                add_assoc_string(return_value, (char*)dump[i], (char*)dump[i+1]);
+            }
+            rd_kafka_conf_dump_free(dump, cntp);
             break;
         default:
             return;
     }
-
-    array_init(return_value);
-
-    for (i = 0; i < cntp; i+=2) {
-        const char *key = dump[i];
-        const char *value = dump[i+1];
-        add_assoc_string(return_value, (char*)key, (char*)value);
-    }
-
-    rd_kafka_conf_dump_free(dump, cntp);
 }
 /* }}} */
 
--- a/conf.c
+++ b/conf.c
@@ -99,7 +99,7 @@ static void kafka_conf_free(zend_object *object) /* {{{ */
             kafka_conf_callbacks_dtor(&intern->cbs);
             break;
         case KAFKA_TOPIC_CONF:
-            if (intern->u.topic_conf) {
+            if (intern->u.topic_conf && !intern->is_borrowed) {
                 rd_kafka_topic_conf_destroy(intern->u.topic_conf);
             }
             break;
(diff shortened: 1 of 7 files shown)


arnaud-lb:
This diverges from librdkafka's behavior and could lead to problems later. Can we expose `rd_kafka_conf_get_default_topic_conf()` and update the documentation of `Conf::dump()` instead?

ralphschindler:
I'm happy to take this in the direction you prefer. Let me give you the options...

Current usage (and this PR) has ergonomic value. The values that go in via `->set()` global or topic, come out via `->dump()`:

```php
// Current API (the bug — topic props silently missing):
$conf = new RdKafka\Conf();
$conf->set('bootstrap.servers', 'localhost:9092');
$conf->set('auto.offset.reset', 'earliest');

$conf->dump(); // ['bootstrap.servers' => 'localhost:9092', ...] auto.offset.reset not here

// What's in this PR (everything in one dump):
$conf = new RdKafka\Conf();
$conf->set('bootstrap.servers', 'localhost:9092');
$conf->set('auto.offset.reset', 'earliest');

$conf->dump(); // ['bootstrap.servers' => 'localhost:9092', 'auto.offset.reset' => 'earliest', ...]
```

To move them to their own separate API call aligns with the `librdkafka` calls (which satisfies our "light weight" wrapper philosophy):

```php
$conf = new RdKafka\Conf();
$conf->set('auto.offset.reset', 'earliest');

$globalDump = $conf->dump();
$topicConf  = $conf->getDefaultTopicConf(); // new method
$topicDump  = $topicConf ? $topicConf->dump() : [];

// Then if they wanted all of them in a single var:
$allSettings = array_merge($globalDump, $topicDump);
```

Worth Noting: The Java implementation which admittedly has less to do with `librdkafka`, doesn't distinguish from global and topic level config values (similar to `->set()` in librdkafka). It also does not have any way of getting values out of the consumer once it's configured, so librdkafka/php-rdkafka is unique here.

Do you want me to work in `->getDefaultTopicConf()`, or do you want to propose a desired API in PHP I can try to work towards?

Thanks for the feedback!

ralphschindler:
Hmmmm...... thinking more about this:

PR #599 adds `Conf::get(string $name): string` which reads a single property back from a Conf object using rd_kafka_conf_get().

PR #602 fixes Conf::dump() to include topic-level properties that set() silently routes to the embedded default_topic_conf.

There seems to be overlap here... Right now:

```php
$conf->set('auto.offset.reset', 'earliest'); // works fine, routes internally

$conf->dump();                               // PR #602: should it appear here?
$conf->get('auto.offset.reset');             // PR #599: throws "unknown property"
                                             // because rd_kafka_conf_get() doesn't
                                             // know about topic-level properties
```

If we do explicit `getDefaultTopicConf()` (no silent merge in dump()) then #599 has the exact same problem in that get() would throw on any topic-level property, forcing callers to write:

```php
$conf->getDefaultTopicConf()->get('auto.offset.reset');
```

...which again requires knowing which tier a property lives in, something `set()` deliberately hides.

The real issue both PRs expose: set() is tier-blind (intentionally), but every read path (get(), dump()) is tier-aware (unintentionally) which feels inconsistent even if it is what is promoted by `librdkafka`s API.

Seems like the two PRs need to be resolved together with a single answer to the question: should reading config mirror the tier-blind convenience of writing it, or should it expose librdkafka's internal split?

I think it makes sense to hash this out and consider a single approach to both of these PRs? I dont think I have a strong preference (I've never really reached for get() or dump() on config values), so I'd love to know what @arnaud-lb and @nick-zh think here...

arnaud-lb:
I would be in favor of adding `Conf::get()` and `Conf::getDefaultTopicConf()` (with the same behavior as `rd_kafka_conf_get()` and `rd_kafka_conf_get_default_topic_conf()`), and not changing the behavior of `Conf::dump()`. This satisfies our "light weight" wrapper philosophy. It's tempting to try to improve the librdkafka API sometimes, but I believe that doing so would make maintenance more difficult in the long run.

We should probably update the documentation of `Conf::set()` to match `rd_kafka_conf_set()`'s:

```
 * @brief Sets a configuration property.
 *
 * \p conf must have been previously created with rd_kafka_conf_new().
 *
 * Fallthrough:
 * Topic-level configuration properties may be set using this interface
 * in which case they are applied on the \c default_topic_conf.
 * If no \c default_topic_conf has been set one will be created.
 * Any sub-sequent rd_kafka_conf_set_default_topic_conf() calls will
 * replace the current default topic configuration.
 *
 * @returns \c rd_kafka_conf_res_t to indicate success or failure.
 * In case of failure \p errstr is updated to contain a human readable
 * error string.
 *
 * @remark Setting properties or values that were disabled at build time due to
 *         missing dependencies will return RD_KAFKA_CONF_INVALID.
```

CONFIGURATION.md indicates which settings are topic-level.

PR #610, opened 2026-05-31
Fix integer overflow on batch_size in ConsumerTopic::consumeBatch()

iliaal:
`consumeBatch()` validated only that batch_size is positive, then called malloc(sizeof(*rkmessages) * batch_size). That multiply overflows size_t for a large batch_size and the result is never NULL-checked, so an oversized batch_size crashes the worker on a NULL write or overruns an undersized buffer once rd_kafka_consume_batch fills it. consumeBatch now rejects sizes that would overflow the allocation (InvalidArgumentException, like the existing partition check) and uses safe_emalloc/efree. Reproduced under ASan.

### DIFF
--- a/tests/consume_batch_invalid_batch_size.phpt
+++ b/tests/consume_batch_invalid_batch_size.phpt
@@ -1,0 +1,23 @@
+--TEST--
+RdKafka\ConsumerTopic::consumeBatch() rejects an out-of-range $batch_size
+--FILE--
+<?php
+
+$conf = new RdKafka\Conf();
+$conf->set('metadata.broker.list', '127.0.0.1:9092');
+
+$consumer = new RdKafka\Consumer($conf);
+$topic = $consumer->newTopic('test');
+
+foreach ([0, -1, PHP_INT_MAX] as $batchSize) {
+    try {
+        $topic->consumeBatch(0, 0, $batchSize);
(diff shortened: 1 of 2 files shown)


ralphschindler:
Reviewed this against our design rules (thin binding, no new abstractions, and the "no user error should cause a crash or memory leak" safety rule from CONTRIBUTING.md) — this is a solid fix for a real bug, and it's fixing exactly the kind of thing that rule exists for. Verified locally: builds clean, new test passes with no broker running.

Two small non-blocking notes, happy to take as a follow-up rather than holding this up:

1. `tests/consume_batch_invalid_batch_size.phpt` hardcodes `127.0.0.1:9092` instead of going through the `--SKIPIF--`/`test_env.php`/`TEST_KAFKA_BROKERS` convention the rest of `tests/` uses. Doesn't matter today since all three cases (`0`, `-1`, `PHP_INT_MAX`) reject before any I/O happens, but it'll bite if this test ever grows a valid-batch_size/success-path case.
2. The `SIZE_MAX` overflow check duplicates what `safe_emalloc` already guards against internally. Worth a one-line comment explaining why it's still there — it's what turns a would-be fatal `safe_emalloc` abort into a catchable `InvalidArgumentException` instead of a hard crash — so a future refactor doesn't "simplify" it away.

Merging as-is.

PR #611, opened 2026-05-31
Free the rd_kafka_error_t returned by the transaction APIs

iliaal:
`rd_kafka_init_transactions()`, begin/commit/abort, and send_offsets_to_transaction return an error object the caller owns and must free with rd_kafka_error_destroy(). The Producer wrappers passed it to create_kafka_error() to build the exception, then dropped it without destroying it, leaking one rd_kafka_error_t per failed transaction call. They now destroy the error after create_kafka_error() reads its fields. Confirmed with ASan: leaked allocations scale 1:1 with failed calls before the change, flat after.

### DIFF
--- a/rdkafka.c
+++ b/rdkafka.c
@@ -843,7 +843,7 @@ PHP_METHOD(RdKafka_Producer, initTransactions)
 {
     kafka_object *intern;
     zend_long timeout_ms;
-    const rd_kafka_error_t *error;
+    rd_kafka_error_t *error;
 
     if (zend_parse_parameters(ZEND_NUM_ARGS(), "l", &timeout_ms) == FAILURE) {
         return;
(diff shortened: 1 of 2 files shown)


ralphschindler:
Thanks for this, Ilia — good catch. Confirmed against the librdkafka headers: all five of these transaction APIs are documented as returning a caller-owned `rd_kafka_error_t*` that must go through `rd_kafka_error_destroy()`, and we were dropping it on every error path. `create_kafka_error()` copies everything it needs out before returning, so destroying right after is safe.

I pushed one small commit on top: the new test hardcoded `127.0.0.1:9092` with no `--SKIPIF--`, so it failed (intermittently, since it races librdkafka's async connection-failure log against the script's own output) for anyone running `make test` without a broker at that exact address. Brought it in line with `init_transaction_not_configured.phpt` — gated on `TEST_KAFKA_BROKERS` via the existing `integration-tests-check.php` helper instead.

Merging now. Appreciate the fix.

PR #612, opened 2026-06-01
fix build with PHP 8.6+

pavetheway91:
https://github.com/php/php-src/commit/7114314c5a96c362b95663f7e7c9184586721f58
https://github.com/php/php-src/commit/df8ce6ddbd713d7e9163dd60d19648e2e4a8bb80

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -912,7 +912,7 @@ void kafka_conf_minit(INIT_FUNC_ARGS)
 {
     handlers = kafka_default_object_handlers;
     handlers.free_obj = kafka_conf_free;
-    handlers.offset = XtOffsetOf(kafka_conf_object, std);
+    handlers.offset = offsetof(kafka_conf_object, std);
 
     ce_kafka_conf = register_class_RdKafka_Conf();
     ce_kafka_conf->create_object = kafka_conf_new;
--- a/consumer_group_metadata.c
+++ b/consumer_group_metadata.c
@@ -279,7 +279,7 @@ void kafka_consumer_group_metadata_minit(INIT_FUNC_ARGS)
     memcpy(&handlers, &std_object_handlers, sizeof(handlers));
     handlers.free_obj = free_object;
     handlers.clone_obj = NULL;
-    handlers.offset = XtOffsetOf(kafka_consumer_group_metadata_object, std);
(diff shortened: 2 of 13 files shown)


pavetheway91:
Just one test (controller_id) fails on 8.6 currently

```
========DIFF========
001- 1
002- 1
003- 1
001+ 0
002+ 0
003+ 0
========DONE========
FAIL Display controller id [tests/controller_id.phpt]
```

edit: Test also fails with PHP85, which leads me to believe that it might fail due to some kind of FreeBSD issue, such as an rather old librdkafka in our package repo.

ralphschindler:
Thanks — verified this against the actual upstream commits rather than just the PR description. Both are genuine, already-merged breaks in php-src, not speculative:

- `XtOffsetOf` → `offsetof`: php/php-src@7114314 deletes the `XtOffsetOf` macro definition outright. It was never more than `#define XtOffsetOf(s_type, field) offsetof(s_type, field)` (a pre-C99 portability shim going back to X11's Xt toolkit), so swapping call sites to `offsetof()` directly is behavior-identical, not a functional change.
- `zval_dtor()` → `zval_ptr_dtor_nogc()`: php/php-src@df8ce6d removes `zval_dtor()`, which was kept only as a compatibility alias for `zval_ptr_dtor_nogc()` (existing since the PHP 7 zval refactor). Same story: no-op substitution.

Grep'd the whole tree — this covers all 13 call sites for both APIs, nothing missed. Built and ran the full suite locally against PHP 8.5.6: clean, 0 failures, including the metadata-object tests that exercise the changed destructor path.

One caveat worth flagging for later: our CI matrix doesn't have a PHP-master job (we do have one for librdkafka-master), so nothing here would have caught this automatically before a user hit it — this PR is a reactive fix, not a systemic guard. Might be worth a follow-up adding that.

On the `controller_id.phpt` failure noted above: confirmed it's unrelated to this PR — the changed code paths here never touch `rd_kafka_controllerid()`, it also failed for you on 8.5 (ruling out an 8.6-specific cause), and it passes cleanly across all 19 jobs in this PR's own CI run. Looks like a stale librdkafka in your FreeBSD package repo, not something to fix here.

Merging — thanks for tracking this down.

PR #613, opened 2026-06-03
Document librdkafka <= 2.8.0 SASL/OAUTHBEARER segfault on process exit

ralphschindler:
## Summary

- Adds a **Known Issues** section to the README documenting a librdkafka bug that causes a segfault (exit 139) during PHP shutdown when using `SASL/OAUTHBEARER` with multiple sequential `KafkaConsumer` instances.

## Background

The crash manifests after all tests/logic complete successfully — PHP shuts down cleanly but the process then segfaults. It is **not** a bug in ext-rdkafka.

**Root cause:** librdkafka bug [confluentinc/librdkafka#4557](https://github.com/confluentinc/librdkafka/pull/4557) — broker background threads are not fully joined before `rd_kafka_destroy()` returns. Orphaned threads access freed memory at process exit.

- Present in all librdkafka versions up to and including **2.8.0**
- Fixed in **2.10.0** (no 2.9.0 release existed)
- Confirmed clean on librdkafka 2.14.1

## Test Plan

- [ ] Verify README renders correctly
- [ ] Confirm the linked librdkafka PR (#4557) is accurate

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,21 @@ while ($producer->getOutQLen() > 0) {
 }
 ```
 
+## Known Issues
+
+### SASL/OAUTHBEARER: process segfault on exit with librdkafka < 2.10.0
+
+When using `sasl.mechanism=OAUTHBEARER` (including `enable.sasl.oauthbearer.unsecure.jwt=true`),
+processes that create and destroy multiple `KafkaConsumer` instances may segfault during PHP
+shutdown (exit code 139) even though all tests or application logic completed successfully.
+
+This is a librdkafka bug ([#4557](https://github.com/confluentinc/librdkafka/pull/4557)) where
+broker background threads are not fully joined before `rd_kafka_destroy()` returns. The orphaned
+threads access freed memory when the process exits.
+
+**Fix:** upgrade librdkafka to **>= 2.10.0**. The bug is present in all librdkafka versions up to
+and including 2.8.0 and was fixed in 2.10.0.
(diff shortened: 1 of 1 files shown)


ralphschindler:
Dug into this before merging and I don't think the root cause here is right. The librdkafka PR I cited (#4557, "Purge brokers no longer reported in metadata") doesn't actually match the description at all — it's about pruning stale broker entries from client-side lists, nothing to do with thread joins, OAUTHBEARER, or process-exit crashes. I couldn't find a real librdkafka issue matching what I described here.

What actually explains the symptom (segfault on shutdown after creating/closing multiple sequential KafkaConsumer instances) is already in our own history: #539/#540/#595 fixed exactly this back in April — `KafkaConsumer::close()` was calling `rd_kafka_consumer_close()` and nulling `rk` without ever calling `rd_kafka_destroy()`, so the free handler's `if (intern->rk)` guard skipped cleanup and the handle leaked, leading to the unclean shutdown / segfault. That's a real ext-rdkafka bug, not a librdkafka one, and it's already fixed on `7.x`.

Since the actual bug is already resolved on `7.x` (the branch this README targets), there's nothing current for this doc to describe — a "Known Issues" entry here would be documenting a problem that doesn't exist for anyone reading this file. Closing without merging rather than publish an unverified root cause.

(#614 tracks the same underlying bug still open on 6.x — no backport planned, closing that out separately.)

PR #618, opened 2026-09-07
Fix getRebalanceProtocol() feature detection

binaryfire:
`KafkaConsumer::getRebalanceProtocol()` requires librdkafka 1.6 or newer, but the source stub wasn’t guarded. Commit `552f353` added the guard directly to the generated `kafka_consumer_arginfo.h`, so regenerating the arginfo could remove it and break builds using librdkafka 1.5.x.

This PR guards the stub, enables the method in Windows builds, and skips its test when the native API isn’t available. It also documents which methods still need librdkafka 1.6 or newer.

The upgrade guide wasn’t included in PECL packages, so this adds it to the package manifest.

While updating `config.w32`, it corrects the version listed for incremental assignment from 1.5 to 1.6. This is only a comment change. The API isn’t in the 1.5.3 header and is present in 1.6.0.

Tested with librdkafka 1.5.3, 1.6.0, and 2.14.1.


ralphschindler:
~~Verified all of this independently before merging:~~

- Reproduced the core bug myself: ran build/gen_stub.php against the current (unpatched) stub and confirmed it silently strips all three HAS_RD_KAFKA_REBALANCE_PROTOCOL guard blocks from the generated arginfo. Your fix regenerates to a byte-identical file except the stub hash, confirming it's complete and correct.
- Confirmed the Windows gap independently: config.w32 never defined HAS_RD_KAFKA_REBALANCE_PROTOCOL at all, so the method has been silently unreachable on every Windows build since it was added. Your fix follows the existing file's own convention exactly.
- Checked the 1.5.0 to 1.6.0 comment correction against librdkafka's actual v1.6.0 release notes (KIP-429 incremental rebalancing lands there, nothing in any 1.5.x changelog). Accurate.
- Confirmed the UPGRADE.md packaging gap against package.xml's own changelog, which already credited adding the file while the manifest never listed it.
- The test SKIPIF pattern matches existing precedent (tests/conf_setDefaultTopicConf8.phpt already uses the same method_exists() idiom).
- Built and ran the full suite locally: clean. CI: 19/19 green.

~~Thanks for catching this, and for the thorough writeup. Merging.~~

ralphschindler:
Sorry I jumped the gun on merging this. Hit the wrong button and was unable to unwind it in `7.x`. Do you mind re-opening this?

binaryfire:
Thanks, no problem. I’ve opened #626 with the changes updated for the 1.6.0 minimum proposed in #624. It removes the 1.6 feature checks instead of fixing them, keeps the `UPGRADE.md` packaging fix, and updates the guide to match.

PR #620, opened 2026-09-07
Fix unconstructed ProducerTopic crash and record conversion

binaryfire:
`ProducerTopic::produce()` and `ProducerTopic::producev()` don't stop when `get_kafka_topic_object()` finds an unconstructed object. This causes a null dereference:

```php
$reflection = new ReflectionClass(RdKafka\ProducerTopic::class);
$topic = $reflection->newInstanceWithoutConstructor();

$topic->produce(0, 0, 'message');
```

Current 7.x exits with status 139. With this change, both producer topic send methods raise the existing `RdKafka\Topic::__construct() has not been called` exception instead of continuing.

This also fixes several record conversion problems:

- Integer header keys are sent as decimal names and no longer stop later headers from being processed. This means `['key']` now sends one header named `0`. Before this change, it sent no headers.
- Header values are converted without changing the input array. If conversion throws, the original exception is kept and no record is queued.
- Header names are passed to librdkafka with their explicit lengths, so embedded null bytes aren't truncated when writing.
- Successful messages with a null payload now keep their Kafka timestamp.

The existing read API still can't read a full header name containing a null byte because `rd_kafka_header_get_all()` doesn't return the name length. The PHPT covers the prefix returned when reading. A separate length-aware consumer confirmed that the full outgoing `nul\0suffix` name is written as `6e 75 6c 00 73 75 66 66 69 78`.

The record conversion tests pass with librdkafka 1.5.3, 1.7.0, 1.9.2, 2.6.0, and 2.14.1. The full 70-test suite passes with 2.6.0 and 2.14.1, and all seven focused tests pass under Valgrind with 2.6.0.

I benchmarked `producev()` before and after this change and didn't find any slowdown.


binaryfire:
Closing this in favor of #628, which includes these fixes and adds ordered message headers.

PR #621, opened 2026-09-07
Fix callback cycle collection and topic ownership

binaryfire:
Some callbacks and parent references aren't visible to PHP's garbage collector, so reference cycles can keep Kafka clients and their threads alive for the lifetime of a worker.

`KafkaConsumerTopic` also didn't keep its consumer alive. Destroying the consumer before a retained topic caused a use-after-free when the topic was accessed or released.

This makes those references visible to the garbage collector and safely invalidates retained topics when their consumer closes.

I ran all 67 tests normally and under Valgrind, with no failures or leaks. I also tested PHP 8.1 with librdkafka 1.7 and PHP 8.3 with librdkafka 2.14.1.

I benchmarked object creation, cycle collection, and topic lifecycles before and after and didn't find a measurable performance regression.


binaryfire:

PR #622, opened 2026-09-07
Prevent KafkaConsumer::close() inside callbacks

binaryfire:
`KafkaConsumer::close()` can destroy the native client while librdkafka is still running a PHP callback. When the callback returns, librdkafka can keep using the freed client and crash or hang.

This tracks nested callbacks and throws `RdKafka\Exception` if `close()` is called before the outer callback finishes. Closing afterward still works, and rebalance callbacks can still assign and revoke partitions normally.

The tests cover nested callbacks, returned object cleanup, rebalance callbacks, and closing normally afterward. They fail against 7.x and pass with this change under Valgrind.

I benchmarked 200,000 delivery callbacks before and after, and there was no performance regression.


binaryfire:

PR #624, opened 2026-09-08
Document the intended librdkafka/PHP support structure

ralphschindler:
## Summary

Raises the 7.x librdkafka floor from 1.5.3 to 1.6.0, updates CI to build/test against that floor, and updates docs to match. Also fixes two stale top-badges and adds a `Supported Versions` section clarifying what 7.x actually requires vs. what CI actually exercises, plus a distro reference table.

## What prompted this

Reviewing #618 surfaced that `getRebalanceProtocol()`/`incrementalAssign()`/`incrementalUnassign()` require librdkafka >= 1.6.0, while this project's CI matrix only builds/tests against v1.7.0 and newer. Nothing between the stated floor (1.5.3) and 1.7.0 was ever built by CI, which is exactly the blind spot that let #618's stub/arginfo guard mismatch land unnoticed. That prompted a closer look at what the 1.5.3 floor and the CI floor actually mean in practice.

## Research (for the record)

- **librdkafka 1.5.3 is upstream-orphaned.** Checked every 1.5.x tag: v1.5.3 (released 2020-12-09) was the last release in that line; development moved straight to 1.6.0. librdkafka doesn't maintain long-term branches per minor version (no SECURITY.md or support-policy doc states one, and there's no backport pattern in the tag history), so 1.5.3 has received zero fixes of any kind in ~6 years.
- **6.x's floor was never raised** — it's still librdkafka >= 0.11.0, unchanged since inception. The 1.5.3 floor is specific to 7.x, set when 7.x was cut as a clean break (package.xml credits it to #559, "Drop unsupported php versions and old librdkafka versions").
- **Distro survey** (via [repology.org](https://repology.org/project/librdkafka/versions)):

  | Distro | librdkafka |
  |---|---|
  | Ubuntu 22.04 (Jammy, supported to 2027) | 1.8.0 |
  | Ubuntu 24.04 (Noble, supported to 2029) | 2.3.0 |
  | Ubuntu 26.04 (Resolute) | 2.13.0 |
  | Debian 11 (Bullseye) | 1.6.0 |
  | Debian 12 (Bookworm) | 2.0.2 |
  | Debian 13 (Trixie) | 2.8.0 |
  | RHEL 8 family (AlmaLinux/Rocky/CentOS Stream 8) | 1.6.1 |
  | RHEL 9 family (AlmaLinux/Rocky/CentOS Stream 9) | 1.6.1 |
  | CentOS Stream 10 | 2.1.1 |
  | Fedora (39–44) | 2.1.1 – 2.14.1 |
  | Ubuntu 20.04 (Focal) | 1.2.1 (below our floor already) |
  | Amazon Linux 2 | 0.11.4 (below our floor already) |
  | CentOS 7 | 0.11.4 (below our floor already) |
  | openSUSE Leap 15.x | 0.11.6 (below our floor already) |

  Nothing currently-relevant ships in the 1.5.3–1.6.0 gap. Every distro below our floor today is already below it regardless of any future decision on our end.
- **Debian 11's stock package (1.6.0) is the one edge case worth tracking**: it clears 1.5.3 but not the 1.7.0 CI floor. Checked Debian's own LTS schedule directly (wiki.debian.org/LTS): Debian 11's free LTS support ended **2026-08-31**. It now only gets updates through Debian's paid ELTS program (2026-09-01 through 2031-06-30), which isn't part of a default `apt` install. So Debian 11's free-support population is already past EOL as of this PR.

## Decision

Raised the floor to **1.6.0** rather than 1.7.0. Per the survey above, nothing currently-relevant sits in the 1.5.3–1.6.0 gap, so 1.6.0 costs nothing, while 1.7.0 would exclude RHEL 8/9's stock 1.6.1 package for no real benefit.

## Changes

- `config.m4`: version-check macro raised from `0x010503ff` to `0x010600ff`; error message now reads "librdkafka version 1.6.0 or greater required."
- `.github/workflows/test.yml`: the floor CI job changed from `v1.7.0` to `v1.6.1` (matching what RHEL 8/9 actually ship, since no distro ships exactly 1.6.0)
- `README.md`: floor text updated to `librdkafka >= 1.6.0`, stale badges fixed, `Supported Versions` section + distro table added
- `.appveyor.yml` (Windows CI) intentionally left at `librdkafka-1.5.3`: confirmed windows.php.net's PECL deps mirror has no prebuilt 1.6.x/1.7.x/1.8.x/1.9.x/2.x package for vs16-x64, only 1.5.3. Raising it there would 404 and break Windows CI — tracked as a follow-up if Windows CI's floor needs to move too.


binaryfire:
One small correction on the Windows archives: the official vs16-x64 listing currently includes librdkafka 1.6.2, 2.12.1, and 2.14.1: https://downloads.php.net/~windows/pecl/deps/. `.appveyor.yml` can therefore move off 1.5.3; I’ve done that in #626, using 2.12.1 to match the Windows release workflow.

Also, current 7.x already calls `rd_kafka_conf_get_default_topic_conf()` without a guard. That symbol was added in 1.6.0, so 1.5.3 can’t build the current source even before the guard cleanup.

Since Debian 11 ships exactly 1.6.0 and configure accepts it, could the floor jobs use v1.6.0 rather than v1.6.1? I built and ran the suite against exact 1.6.0 in #626.

One test note: `conf_callbacks_integration.phpt` is already flaky on librdkafka 1.x because the default logger can append an extra stderr line to its exact expectation. It sets neither `setLogCb()` nor `log_level`; adding the same no-op log callback used by `kafka_consumer_get_rebalance_protocol.phpt` would make that test deterministic. I kept this unrelated test change out of #626.

PR #626, opened 2026-09-09
Remove obsolete librdkafka 1.6 feature checks

binaryfire:
This builds on #624, which raises the minimum librdkafka version to 1.6.0. `KafkaConsumer::getRebalanceProtocol()`, incremental assignment, and the full `ConsumerGroupMetadata` constructor all use APIs available in that version, so their feature checks are no longer needed.

This removes those checks and makes the APIs unconditional. It also updates the dormant AppVeyor dependency to the same 2.12.1 archive used by the Windows release workflow, updates `UPGRADE.md` for the new minimum, and includes that guide in PECL packages.

The old incremental assignment check passed two symbols to one `AC_CHECK_LIB`, so autoconf only tested `rd_kafka_incremental_unassign`. With `-Werror`, the generated check also fails and silently removes both methods.

Tested with librdkafka 1.6.0 and 2.14.1. Both built cleanly with `-Wall -Werror`; 1.6.0 passed 62 tests with 3 expected skips for the 2.8-only metadata getters, and 2.14.1 passed all 65. PECL package validation and an extracted archive build also passed.


binaryfire:

